Issue #1, opened 2016-12-20
chandra_repro with badpixel=no


kglotfelty:
The verbose output from chandra_repro when the `badpixel=no` is wrong

```bash
% chandra_repro 333 out='' badpixel=no clob+

Processing input directory '/data/sdsreg/thread_output/CIAO4.9/acisbadpixels/333'

WARNING: badpixel=no, so archived bpix1.fits file will be used.
```

should be __will not__ be used.

PR #14, opened 2017-03-06
chandra_repro fix typo when badpixel=no

kglotfelty:
Add a missing "not" to verbose output when run with badpixel=no.


kglotfelty:
This is issue #1

Issue #7, opened 2017-02-02
rfe: add support for typeI pha files in mktgresp


kglotfelty:
From a DPH email to ciaoscripts --

He's trying to use the old fullgarf script to create ARF but running into issues with FITS format.  I  suggesetd mktgresp but it doesn't work with type I pha files.  

mktgresp only needs the order, part, scrid and position from the pha file.  It's currently using crates to get them -- expecting them to be columns.  In a type I file, they will be keywords.   It's easy to make the code try/except to read them from both pha flavors.

PR #18, opened 2017-03-06
mktgresp support typeI files, tgsplit include REGION extension

kglotfelty:
Ref: #7 

updated mktgresp to allow for type I files; however, mkgrmf requires the REGION extension which are not being copied by tgsplit.  So this also updates tgsplit to append the REGION extension.   This is harder than it seems since going through crates, the region block is dropped so I need to change several places to the dataset's write method


kglotfelty:
new test 08.main in mktgresp regression test adds a case where tgsplit is run to split hrc/letg dataset and then runs mktgresp twice to create response files.

Issue #17, opened 2017-03-06
combine grating spectrum : remove stat_err column


kglotfelty:
Currently c_g_s copies the `STAT_ERR` column from one of the inputs "randomly" (whichever is in dictionary first).  Some tools use these values which are now wrong when more than 1 file is combined.

Since what's there is definitely wrong, we can easily just remove the column.  Need to also check the OGIP standard to see if we need to change header to let s/w know whether to expect column or not.

PR #19, opened 2017-03-06
update combine_spectra to recompute the stat_err column and set POISSERR=True

kglotfelty:
I updated combine spectra (called by combine_grating_spectra) to drop the stat_err column and to se the POISSERR keyword=True.  Ref #17 

All the c_g_s tests are affects but none of the combine_spectra tests themselves are.

specextract with combine=yes is TBD.

### DIFF
--- a/ciao-4.9/contrib/bin/combine_spectra
+++ b/ciao-4.9/contrib/bin/combine_spectra
@@ -20,7 +20,7 @@ from __future__ import print_function
 
 
 toolname = "combine_spectra"
-__revision__ = "13 Sep 2016"
+__revision__ = "06 March 2017"
 
 import sys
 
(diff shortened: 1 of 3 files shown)


DougBurke:
I assume that if we are going to make this change then we may as well make it for both `combine=no` and `combine=yes`?

kglotfelty:
it's actually a noop for specextract.

specextract runs dmextract with defaults (does a punlearn).  The default is  `error=gaussian` which -- for obscure historical reasons, means that it doesn't output the  stat_err column and already sets POISSERR=T .

kglotfelty:
Based on feedback from Dave, we are OK to recompute using Gehrel's since that's essentially what the fitting apps do when there is no stat_err column.  I added an extra message too just to let users know.

This way the structure of the files is unchanged and the values are reasonable.

PR #33, opened 2017-04-03
Seeding Changes files for  4.9.2

kglotfelty:
Added notes from the pull requests for the 4.9.2 (assuming everything is going in).

PR #36, opened 2017-04-06
change log for the contrib 4.9.2 release

nplee:
initial list of changes for the contrib 4.9.2 release based on pull request comments and issues discussion.


DougBurke:
@nplee - can you look through the notes @kglotfelty added in #33 and add any you've missed here? Or are they essentially the same? (this PR has a slightly-cleaner history so is the one I'd prefer to merge).

nplee:
After review, they're essentially the same.  #33 uses bullet points and is terser than #36.

PR #37, opened 2017-04-07
regenerated runtool.py file using the mk_runtool.py script

nplee:
only difference is in comment about the new random parameter


DougBurke:
@nplee can you add a "Changed in version 4.9.2" section to the runtool xml file? Ta - Doug

nplee:
@DougBurke just added a note on 4.9.2 in the xml file; but there's a really large gap since the last update of the file (contrib 4.6.7), which I'll work on filling in.

DougBurke:
@nplee ta. We must have got out of the habit of updating this help file!

PR #40, opened 2017-05-15
updated to use ftp://cda.cfa.harvard.edu as the default FTP 

kglotfelty:
with DS10.6, ftp://cdaftp.cfa.harvard.edu will be decommissioned.  This changes the default FTP site to use 'cda' instead of 'cdaftp' in 
- download_chandra_obsid
- find_chandra_obsid
- download_obsid_caldb


DougBurke:
LGTM

PR #41, opened 2017-05-22
Python 3.5 fixes for deflare and lightcurve

DougBurke:
Use of six.moves.input rather than raw_input and change to
NamedTemporaryFile rather than mkstemp from tempfile, along
with explicit mode setting, to avoid byte/str confusion.


DougBurke:
Oops - forgot the documentation. BRB.

DougBurke:
Doc updates added in #42.

Issue #52, opened 2017-06-23
bug: ciao_contrib/region/fov.py syntax error


kglotfelty:
Hi,

When the FOV file that `skyfov` outputs is _bad_ -- ie using the `marxasp` output with a constant `ROLL` value, we try to report the rows with problems.  There is a syntax error in the error reporting code:

```diff 
diff --git a/ciao-4.9/contrib/lib/python2.7/site-packages/ciao_contrib/region/fov.py b/ciao-4.9/contrib/lib/python2.7/site-packages/ciao_contrib/region/fov.py
index 46c7cd0..8b47e01 100644
--- a/ciao-4.9/contrib/lib/python2.7/site-packages/ciao_contrib/region/fov.py
+++ b/ciao-4.9/contrib/lib/python2.7/site-packages/ciao_contrib/region/fov.py
@@ -596,7 +596,7 @@ class FOVRegion:
         if len(idx) != cr.get_nrows():
             idx, = np.where(shapes != "Polygon")
             bad = [str(d)
-                   for d in cr.get_column(cr, "CCD_ID").values[idx]]
+                   for d in cr.get_column("CCD_ID").values[idx]]
             wmsg = "WARNING: Unable to read regions for CCDs " + \
                 "{} in obsid {}".format(", ".join(bad), self.obsid)
             v1(wmsg)
```

This generates an exception in `fluximage` (and probably other scripts that are using this routine).
```bash
% fluximage testmarx.fits testmarx/ asolfile=asol.fits badpixfile=NONE maskfile=NONE
...
# fluximage (12 September 2016): ERROR get_column() takes exactly 2 arguments (3 given)
```

Files are available in `/data/lenin2/Bugs/SkyFov/StaticRoll`

DougBurke:
Should be an easy fix. We obviously need more tests of the error handling
routines ...

DougBurke:
Closed by #54

PR #54, opened 2017-06-27
Fix issue #52

DougBurke:
Correct the syntax of the get_column call, as well as remove an un-needed
import.


DougBurke:
I also did a quick manual/visual check of our use of `get_column` and it looks like this is the only place where this was a problem (not 100% guaranteed to be complete :-).

PR #57, opened 2017-07-07
Fix #50 

### LINKED ISSUE #50: mktgresp: problems w/ asphist cleanup in Python 3
mktgresp uses `map()` to clean up the temporary aspect histogram output files. In Python3 this isn't working as expected; at least on osxElCap in CIAO 4.10b1.  This should be re-written to avoid the p2/p3-isms.

kglotfelty:
This is the problem with mktgresp not cleaning up temporary aspect histogram files under Python3.

Regression tests pass; in particular test 08.MAIN which runs mktgresp twice.  Before I would get warnings about unable to clobber existing files during the 2nd run; this fixes that.

### DIFF
--- a/ciao-4.9/contrib/bin/mktgresp
+++ b/ciao-4.9/contrib/bin/mktgresp
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # 
-# Copyright (C) 2013,2016 Smithsonian Astrophysical Observatory
+# Copyright (C) 2013,2016,2017 Smithsonian Astrophysical Observatory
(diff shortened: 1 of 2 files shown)


DougBurke:
Can you add a note to the changelog to act as a seed/reminder for Nick?

kglotfelty:
I doubt it without fouling something up ... took me 2+ hrs to get this to committed cleanly

nplee:
I've taken note of this one on my list, so it'll make it into my changelog edits.

PR #60, opened 2017-07-11
Fix #53

### LINKED ISSUE #53: srcflux using dmmerge to combine temporary files
Based on email from Dong-Woo (Ref: "Fwd: Re: source pipe" 6/26/2017) 

`srcflux` runs `dmextract` to compute the PSF fraction in the source and background regions for each source individually.  These individual output files are then combined into a single table using `dmmerge`.

However, `dmmerge` doesn't just append the rows, it also merges the subspace.  `dmextract` adds the region subspace into the output file.  Thus when `dmmerge` runs, as in DWK's example with 450+ sources, the region intersection is extremely messy; and ultimately useless since we only care about the PSF fractions.  In this case, the merge ends up exhausting system memory (though I'm somewhat surprised why) and the script fails.

We should be able to do something like tack on a `[subspace -sky]` to the files being `dmmerge`d together.  My off-to-the-side test showed that this ran to completion.  I also went ahead and disable the header lookup table since we don't care about the header in this case either.

But I want to see if there are any bad side effects of doing this.  Also I'm wondering if just opening the files via crates would work and appending the rows together that was works -- or will we hit a system limit on number of open file handles (which I'm thinking is what sent me down the `dmmerge` route to begin with).

Note:  There are other places in the script where dmmerge is used in a similar way to collect temp files before adding them to the final output table.  Those need to be reviewed too.

kglotfelty:
Dong-Woo ran into a problem when using a source list as input with >450 sources ; they are inputting their own PSF files and their own regions. 

I use `dmextract` to extract the PSF fractions from the files, individually for each source; the `dmextract` output has the region in the subspace.  Then to combine those I use `dmmerge`. The problem they were finding was that `dmmerge` was running the machine out of memory as it was trying to merge the subspaces.  It was also wasting a lot of time (hours) trying to do this [the input regions are mildly complicated].

Since we don't need the subspace, just the values, the easiest thing is to just remove the subspace from the `dmextract` output as it's being input to `dmmerge`.  There are a couple other places where I do something similar so those other instances of `dmmerge` are also updated.

This has no effect on the output; just the performance.

Regression tests pass.  Dong-Woo's dataset ran to completion in a reasonable 1.5hrs.

### DIFF
--- a/ciao-4.9/contrib/Changes.CIAO_scripts
+++ b/ciao-4.9/contrib/Changes.CIAO_scripts
@@ -15,6 +15,12 @@
       processed. This only changes the error message, and not how
       the tools run.
 
+    srcflux
+
+      Added an internal optimization for the case where a large number of 
+      positions are used, especially when using psfmethod=psffile. 
+      There is no change to the output.
+
     chandra_repro
 
       Fixed internal problem when the observation being processed
--- a/ciao-4.9/contrib/bin/srcflux
+++ b/ciao-4.9/contrib/bin/srcflux
@@ -1,7 +1,7 @@
 #! /usr/bin/env python
 
 #
-# Copyright (C) 2013-2016 Smithsonian Astrophysical Observatory
+# Copyright (C) 2013-2017 Smithsonian Astrophysical Observatory
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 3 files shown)


DougBurke:
I'll rebase for the comment changes.

DougBurke:
Rebased. Original tip was 77e6082

PR #56, opened 2017-06-29
Fix #49

DougBurke:
This adds axbary, evalpos, mean_energy_map, pileup_map, and wavdetect
to the list of tools run with CIAOToolDirect rather than via the
parameter file (using the @@ syntax).

It also adds a minor clean up to the add_tool_history routine, to avoid
clobbering the global dmhistory setting.

PR #63, opened 2017-07-19
Fix #39

### LINKED ISSUE #39: doc: add ASCDS_WORK_PATH info to runtool help file
From
 https://icxc.harvard.edu/pipe/ciaodoc/2015/0102.html

suggested that we add info into the runtool help file re: setting ASCDS_WORK_PATH to change temp dir.

DougBurke:
Add documentation about the use of ASCDS_WORK_PATH to the documentation (mostly ahelp as it is already mentioned in the Python docstrings).


DougBurke:
About to rebase as #56 has landed. Tip was b2e3ce9

PR #73, opened 2017-08-11
multi-chip subspace component support

nplee:
this updates the handling of reference images where the subspace component may contain more than one CCD.  If the reference image from a pseudo-events file, it will have a subspace component for each CCD, but often, the clipped counts image produced by fluximage will have a multiple CCDs in a single subspace component.  Not accounting for this can lead to index bounds issues causing the script to break.

### DIFF
--- a/ciao-4.9/contrib/bin/blanksky_image
+++ b/ciao-4.9/contrib/bin/blanksky_image
@@ -1,8 +1,30 @@
 #!/usr/bin/env python
 
+#
+# Copyright (C) 2010-2012, 2013, 2014, 2015, 2016, 2017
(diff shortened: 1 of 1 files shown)


DougBurke:
@nplee - I've merged this. Does the ahelp file for `blanksky_image` need updating, or was this an error found by you rather than from helpdesk?

nplee:
the ahelp doesn't need to be altered, this originated from helpdesk. The script originally intended to handle this case, but I didn't realize how taking out the time subspace in the event file could impact the CCD subspace components, so approached the problem in a more general way to do what it's suppose to do.

PR #82, opened 2017-11-06
added checks for interleaved observations so they're mergeable with merge_obs

nplee:
added checks for interleaved observations so they aren't treated as as multi-obi observations, since they will be removed from merging due to changes between CIAO 4.6 and 4.7, affecting merge_obs.


DougBurke:
@nplee - will we need to add a note to `merge_obs.xml` to say something like "handling of interleaved observations has been improved"? (not entirely sure what this fix improves)

nplee:
I wouldn't really consider this an "improvement" so much as a "bug" due to not making changes in the script by differing multi-ObI and interleaved observations which had to be accounted for starting in CIAO 4.7.  Some extra keyword checking was added to this update so the script should behave as it's intended to, so I don't think this requires an ahelp update.

PR #87, opened 2017-12-12
fixes problems with spectrum if 'channel' parameter not default

nplee:
This change adds the PI channel range/binning to use if the "channel" parameter is specified other than the default.  If "channel" is set to some other value than default, the current behavior is to have the PI file to use channels 1-1024 rather than the ones used to create the RMF, which will cause issues when read into Sherpa.


DougBurke:
@nplee - does this need an update to the ahelp file?

nplee:
I think I can massage the parameter description a little bit.

PR #101, opened 2018-03-06
added note that the channel parameter will also be applied to the PI …

nplee:
…spectrum if the default value is not used in the ahelp file


DougBurke:
About to rebase. Tip was 5fa62f9

DougBurke:
Actually, I ended up creating a new pr that replaces this one. See #105

PR #114, opened 2018-03-29
srcflux deal with blank/NULL regions as before

kglotfelty:
Before ciao-4.10, a 0 row FITS region was read in as "point(0,0)" .  This happens in srcflux when the regions doesn't overlap with the FOV (a null intersection).  

In CIAO 4.10, a 0 row FITS region file is "" (blank|NULL); but srcflux is setup to look for the point(0,0).  

Without this change, there is a ValueError exception thrown when the source stack is expanded because expanding a stack file with 0 non-blank rows is an error.

### DIFF
--- a/ciao-4.10/contrib/bin/srcflux
+++ b/ciao-4.10/contrib/bin/srcflux
@@ -313,7 +313,13 @@ def save_regions( myparams, regions, flavor ):
     """    
     from ciao_contrib.runtool import dmmakereg
     from tempfile import NamedTemporaryFile    
-    ss = stk.build(regions)
+
+    # CIAO 4.10 changes to NULL regions no longer creates a fake
+    # "point(0,0)" -- I'm using that so we'll put it back.
+    try:
+        ss = stk.build(regions)
+    except ValueError as ve:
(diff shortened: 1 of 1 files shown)


ciao-4.10/contrib/bin/srcflux
```diff
@@ -313,7 +313,13 @@ def save_regions( myparams, regions, flavor ):
     """    
     from ciao_contrib.runtool import dmmakereg
     from tempfile import NamedTemporaryFile    
-    ss = stk.build(regions)
+
+    # CIAO 4.10 changes to NULL regions no longer creates a fake
+    # "point(0,0)" -- I'm using that so we'll put it back.
+    try:
+        ss = stk.build(regions)
+    except ValueError as ve:
```
DougBurke: In general I'm a bit suspicious of code catching such a generic error - i.e. what happens if there's an error thrown because of a case other than "is empty" - e.g. perhaps the file is corrupted or been eaten by a gru. It's not obvious from a whole 10 seconds of thinking about it what a cleaner way to handle this case is though.

I guess the fact you read from the file below catches the case of "there's something borked with the file" case.

PR #117, opened 2018-03-30
Use ROLL_PNT in preference to ROLL_NOM and pixlib for plate scales (fix #23)

### LINKED ISSUE #23: make_psf_asymmetry_region NOM vs PNT
(Posting here from ciaoscript email last year)

Just a quick note ... make_psf_asymmetry_region is using the ROLL_NOM keyword instead of the ROLL_PNT keyword.  _PNT should be used, if it exists.

[boiler-plate : they are diff for multi-obi and reprojected/astrometric-corrected datasets; however PNT is purposefully dropped from L2 file, and is '0' in the asol files.]

Also it's using hard-code plate scales; I don't expect they'll change, but would be better long-term to get via pixlib or the caldb file.

DougBurke:
This does change the plate scale used for HRC data from 0.13175 to 0.1318, which results in a ~0.05% reduction in HRC radii, which has been mentioned in the documentation.

### DIFF
--- a/ciao-4.10/contrib/Changes.CIAO_scripts
+++ b/ciao-4.10/contrib/Changes.CIAO_scripts
@@ -49,6 +49,14 @@
     
         Force script to use bash shell to fix issue on Ubuntu which 
         defaults to using 'dash'.
+
+    make_psf_asymmetry_region
+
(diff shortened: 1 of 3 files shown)


DougBurke:
@kglotfelty does this look sensible?

PR #120, opened 2018-04-13
revised change log update for CIAO 4.10

nplee:
revised using upstream/master for CIAO 4.10; references to CIAO 4.9 updates removed.


DougBurke:
Thanks for the re-work. No idea where these extra commits are coming from! Life is too short to wonder though.

PR #125, opened 2018-05-14
enable 'rmffile' for HRC observations

nplee:
this is to support the 'rmffile' parameter for HRC zeroth-order


DougBurke:
It is probably worth saying something like "It is expected that rmffile is left at CALDB for HRC data."

DougBurke:
I ended up rebasing this to avoid the merge, so this is being closed. See #126 for the updated version.

PR #131, opened 2018-06-25
Support current view in search CSC scripts

kglotfelty:
These change fold in previous changes to allow the `search_csc` and `obsids_search_csc` scripts to access the current, aka `csc2` , tables **and** data products.

Both scripts now support

 `catalog=cur`

which allows access to the current database.  For example:

```bash
search_csc 3c273 out=3c273.tsv radius=0.1 radunit=arcmin download=all \
      file=regimg,regexp,regevt,reg,psf,arf,rmf,pha,lc clob+  cat=cur
```

> Note:  There will be some work when `csc2` is _released_ to update the parameter file and update a few `cur` conditionals.

I wanted to get this pull request in sooner-rather-than-later so folks could test with it.  Once we have a release plan, I'll update the ahelp files.

### DIFF
--- a/ciao-4.10/contrib/bin/obsid_search_csc
+++ b/ciao-4.10/contrib/bin/obsid_search_csc
@@ -1,26 +1,26 @@
 #!/usr/bin/env python
-#
-# Copyright (C) 2013,2016,2018 Smithsonian Astrophysical Observatory
-#
+# 
+# Copyright (C) 2013,2016 Smithsonian Astrophysical Observatory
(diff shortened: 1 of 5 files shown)


DougBurke:
If we merge this code into master then users can't actually use this because the `.par` files need to be updated to support setting `catalog=cur` (I think)?

kglotfelty:
Correct.  They could hack the enumeration ... but nobody’s going to do that.

Kenny

PR #132, opened 2018-06-28
Fix broken map() in mktgresp which fails to iterate over both HEG and

kglotfelty:
… MEG parts

From HD ticket 21345, the MEG responses are not created in the py3 build.

Users can set `orders=` (blank string) and it works.

### DIFF
--- a/ciao-4.10/contrib/bin/mktgresp
+++ b/ciao-4.10/contrib/bin/mktgresp
@@ -119,7 +119,7 @@ def parse_orderlist( orderlist, parts, ids, xx, yy ):
         return None
 
     try:
-        ol = map(int, ss)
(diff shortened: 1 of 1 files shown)


DougBurke:
Hmmm, why are there two identical-looking commits here?

Also, can we have doc updates too (both to the ahelp file and the changes file)? Ta.

kglotfelty:
not quite identical "int" vs "list" .... ie, I fouled up the 1st one after another round of which-branch-am-I-on-anyway-bad-merge-failed-rebase.

I was holding off on docs until we had a date/plan/month to avoid multiple edits to same file

DougBurke:
I'm going to merge and we can work on docs "soon".

PR #149, opened 2018-10-23
blanksky

nplee:
this introduces sub-array support for the 'blanksky' script and releasing the 'blanksky_sample' script and associated files into the wild.


ciao-4.10/contrib/bin/blanksky
```diff
@@ -137,24 +138,51 @@ def caldb_blanksky(detector):
...
+
+        try:
+            obs_time = evt_keywords["LIVETIM{}".format(chip)] 
+        except KeyError:
+            try:
+                obs_time = evt_keywords["LIVTIME{}".format(chip)]
+            except KeyError:
+                obs_time = evt_keywords["LIVETIME"]
+
+        # if evt_keywords["DATAMODE"] == "GRADED":
```
DougBurke: Nick, could we have a comment here saying why the following has been commented out (i.e. why it is worth carrying around)?

nplee: I've removed the commented out code, as they are unnecessary and just adds confusion.

DougBurke:
This has been merged in #155 (which just cleaned up the git history)

PR #168, opened 2018-11-19
chandra_repro update to keep all columns for HRC

### LINKED ISSUE #146: RFE:  nix cols filter for HRC L2 event files
https://github.com/cxcsds/ciao-contrib/blob/4b320f51cb5d43a347916b206726f509103d961c/ciao-4.10/contrib/bin/chandra_repro#L1904

There may be a new eventdef coming in 4.11 (adding `samp` ?) so we'll update this, then.

kglotfelty:
This fixes #146 .

(acis is already keeping all cols)

### DIFF
--- a/ciao-4.11/contrib/Changes.CIAO_scripts
+++ b/ciao-4.11/contrib/Changes.CIAO_scripts
@@ -1,6 +1,14 @@
 
 ## 4.11.1 - December 2018 ##
 
+Update scripts
+
+  chandra_repro
+  
+    The script has been updated for HRC to keep all the standard 
+    columns defined in the hrc_process_events stdlev1 event-definition 
+    parameter.  In CIAO 4.11 this now also includes the SAMP values.
+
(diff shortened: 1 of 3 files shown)


DougBurke:
I assume the old logic (that lead to columns being removed) was there to reduce the file size. I assume that the saving isn't that huge (in the grand scheme of things) so we just keep things in as it makes downstream processing easier (rather than trying to support optional filtering, where we either decide whether to remove columns automatically or via a user choice). 

In other words, we could tell users that they can remove these excess columns if they don't need them / the files are too large, in the documentation. If we felt it was helpful / worthwhile.

kglotfelty:
Correct.  Keeping the columns should never hurt.  [Unless someone does something wrong by accessing columns by number instead of by name -- then they get what they deserve.]

Byte wise this does increase the files more than a little but big-picture, just a drop in the bucket.

PR #169, opened 2018-11-20
Create PSF maps in fluximage, flux_obs, and merge_obs

DougBurke:
Adds the ability to create a PSF map for the observation to fluximage. This is done with a single parameter, `psfecf`, which when left at its default value (`INDEF`) means that no PSF map is created. When `psfecf` is set to a value `[0,1)` then a `<band>.psfmap` is created per band. The energy used for the band is the same as for the instrument map (e.g. 2.3 keV for broad band).

The `flux_obs` and `merge_obs` script pass the `psfecf` parameter on to the `fluximage` call, and have a new `psfmerge` parameter to control how the PSF maps should be combined (e.g. min value, exposure-map weighting). The default is to use the minimum pixel value.

The PR now supports

- [x] filtering the PSF map by the FOV dile for the observation, to mark unobserved pixels as 0/filtered out 
- [x] support for spectrally-weighted bands
- [x] extend to `flux_obs` and `merge_obs`
- [x] fix the psfmap names to include `_thresh` when thresholding is applied

There has been limited testing.


DougBurke:
I know, he thinks to himself, I could implement the filter using the exposure map converted to a MASK rather than use skyfov.

kglotfelty:
You may want to add opt all just in case the expmap pulls away from the edge

DougBurke:
The images should cover the same x/y grid by construction; if they don't
then the code is buggggie.

kglotfelty:
I’m just not sure if this is the threshold’ed (?) file.  I think the pixelmask now bounds to the interior, non-zero part .. but that’s one of those things that kept changing 

Kenny

PR #185, opened 2018-11-29
Change ardlibqual to hidden and fix a typo

DougBurke:
This was originally a fix for a small typo in the readme, but then I noticed that the new `ardlibqual` parameter to `fullgarf` was listed as automatic (`a`) rather than hidden (`h`).


DougBurke:
@kglotfelty - I think this parameter file change makes sense, but can you check? Also, @hamogu does it make sense to you?

PR #186, opened 2018-11-29
chandra_repro: do not write out a second HISTORY block (fix #175)

### LINKED ISSUE #175: update clear_acis_status_bits to use new HISTORY module
https://github.com/cxcsds/ciao-contrib/blob/fe1fe329ae4bc71da9f3e855a44dad5158c65f50/ciao-4.11/contrib/lib/python3.5/site-packages/ciao_contrib/cxcdm_wrapper.py#L633-L635

This is related to #174 .  We can use the new in CIAO 4.11 `history` module to generate standard history records w/o a parameter file.

DougBurke:
The clear acis status bits functionality - called by chandra_repro - added
a HISTORY record in non-standard format, which was causing some confusion.
The change is to make this header addition optional, and for chandra_repro
not to make the change.

One option would be for chandra_repro to call clear_acis_status_bits
with toolname set to 'acis_clear_status_bits', but that is a bit risky
(when/if the script gets removed or renamed).

Note: I did not add any documentation about this change to chandra_repro, as I don't think it's worth it.

### DIFF
--- a/ciao-4.11/contrib/bin/chandra_repro
+++ b/ciao-4.11/contrib/bin/chandra_repro
@@ -1009,7 +1009,7 @@ def reset_acis_status(params):
     v5("Filename for reset event file: " + reset_file)
 
     v1("Resetting afterglow status bits in evt1.fits file...")
-    cdw.clear_acis_status_bits(reset_file, "chandra_repro")
+    cdw.clear_acis_status_bits(reset_file)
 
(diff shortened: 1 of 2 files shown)


DougBurke:
@kglotfelty - this look okay (I took the opportunity to change chandra_repro to use the new functionality, but haven't tested it).

kglotfelty:
This looks OK.  I ran regression tests and looks good.

PR #188, opened 2018-11-29
splitroi update to restore point(0,0) behavior

kglotfelty:
In CIAO 4.10, FITS regions with 0 rows are now returning
a region with 0 shapes to which regRegionString() 
returns a blank string.  stk_expand on blank lines
are lost (multiple-stack separators are collapsed into 1)
this may produce different number of regions in the src
and bg outputs.

This workaround restores the point(0,0) which was what
he DM/regionlib did before CIAO 4.10. 

Ref: CXC Helpdesk ticket 21625 where user running `srcflux` (which runs `splitroi`) ran into this very problem : a mismatch between number of src and background regions.

### DIFF
--- a/ciao-4.11/contrib/Changes.CIAO_scripts
+++ b/ciao-4.11/contrib/Changes.CIAO_scripts
@@ -30,6 +30,11 @@ Updated scripts
     internal changes to use python scripts to perform sherpa fits
     rather than using sherpa IPython wrapper.  
 
+  splitroi
+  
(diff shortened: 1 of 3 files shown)


DougBurke:
I'm slightly worried that the fix is actually "too" good, in that even when (if?) we have this fixed upstream in CIAO, the `getGetNoShapes` call could still return 0, and so we won't benefit from the fix here. However, this is worrying about unspecified future behavior that may never be, so I'm not going to let that stop me from merging this. Ta.

PR #195, opened 2018-12-21
Try and support CERTIFICATE_VERIFY_FAILED errors

DougBurke:
It is not clear why we are getting these messages, so try and fall
through to curl/wget if we do get them. Seems to be a macOS issue
at present.

Should the csccli routines be updated to use this "downloadutils"
module? Note that at the moment the csccli access doesn't need
https support I believe.


DougBurke:
If this code works I'll squash it into one commit before merging.

nplee:
I was just able to run a query when it was previously failing.

PR #198, opened 2019-01-25
Blanksky image chips

nplee:
modified script to better handle input blanksky event files that have chip filters

### DIFF
--- a/ciao-4.11/contrib/bin/blanksky_image
+++ b/ciao-4.11/contrib/bin/blanksky_image
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 #
-# Copyright (C) 2010-2012, 2013, 2014, 2015, 2016, 2017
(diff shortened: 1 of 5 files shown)


DougBurke:
Just about to do some clean up of this PR

DougBurke:
And some more cleanup, to remove the addition/removal of specextract2.

ciao-4.11/contrib/bin/blanksky_image
```diff
@@ -216,6 +224,11 @@ def bkg_img(imgfile,bkg,outfile,dmfilter,dettype,chips,tmpdir,verbose):
 
     cr.__del__()
```
DougBurke: I'm going to remove this - the call to `cr.__del__()` - as it should no longer be needed (I believe it's a hold-over from a crates bug that was fixed a while ago).

DougBurke:
I've added documentation (minimal), removed the `_PNT` keywords as well as the `_AVG` ones, and changed `cr.__del__()` to `cr = None`.

PR #201, opened 2019-02-04
correct_periscope_drift now uses Sherpa to create plots (support matplotlib, fix #202)

### LINKED ISSUE #202: correct_periscope_drift when you have chosen pylab as Sherpa plotting backend
The `correct_periscope_drift` script creates PNG plots as part of its run. Two are created with ChIPS commands directly, and 2 via Sherpa plot commands. This causes problems for any user who has set their `~/.sherpa.rc` file to use the `pylab` backend for Sherpa plots. When I tried it the script failed with an inscrutable error message from ChIPS about a missing window (or some such).

I claim that #201 avoids this (it does ugly things to ignore the users Sherpa configuration), but it needs testing by people other than me with both the default `~/.sherpa.rc` file and one in which `pylab` is chosen for the `plot_pkg` setting.

DougBurke:
# Latest version

The code has been re-worked so that it uses the Sherpa plot routines for creating the plots (which means the plots are slightly different to the ones shown below). This means that the script will now use the `plot_pkg` setting in `~/.sherpa.rc` to determine whether to use ChIPS or Matplotlib.

I have also tweaked the screen output to be a bit-more emphatic about reviewing the output plots (since they will not appear briefly) when Matplotlib is in use.

## Question

We could keep the plotting changes but always force matplotlib.

# Original text

This should now allow users to run correct_periscope_drift when there
is no X display or over a remote connection (both problems with ChIPS),
and on recent macOS versions (where ChIPS is known to be problematic).

The plots are now larger, in color, and will not appear briefly as an
X window when the script is run. The plot files are now over-written
even if clobber is set.

The conversion to use matplotlib uses an "interesting" technique to
work around the use of Sherpa's plot_fit command. What I should have
done was to manually create the equivalent to plot_fit with direct
matplotlib calls, but I instead have decided to see if I can hack
Sherpa's ploting routines to ignore the user's plot_pkg setting and
always use the Sherpa matplotlib backend.

Here's example plots:

corr_fit_zag
corr_data_zag

### DIFF
--- a/ciao-4.11/contrib/share/doc/xml/correct_periscope_drift.xml
+++ b/ciao-4.11/contrib/share/doc/xml/correct_periscope_drift.xml
@@ -2,39 +2,50 @@
 <!DOCTYPE cxchelptopics SYSTEM "CXCHelp.dtd">
 <cxchelptopics>
   <ENTRY key="correct_periscope_drift" context="tools"
-         refkeywords="aspect solution motion source coordinates astrometry alignment
-         correction temporal drift sub-arcsec sub-pixel subarcsec subpixel thermal
-         periscope alignment psf"
+         refkeywords="aspect solution motion source coordinates astrometry
+		      alignment correct correction temporal drift sub-arcsec
+		      sub-pixel subarcsec subpixel thermal aimpoint accurate
+		      accuracy periscope alignment psf"
          seealsogroups="" displayseealsogroups="pixtools asptools">
 
     <SYNOPSIS>
-        Correct a Chandra aspect solution for temporal drift calculated from a fit of the
-        drift in the sky positions of the events in a supplied circle region.
+        Correct a Chandra aspect solution for temporal drift
+        calculated from a fit of the drift in the sky positions of the
+        events in a supplied circle region.
     </SYNOPSIS>
 
     <DESC>
       <PARA>
-        Thermal cycling on the spacecraft can result in an apparent temporal drift of the sky
-        position of an X-ray source during an observation.  This appears as a drift of up to about
-        0.5 arcsec in X-ray event sky X, Y coordinates over time.  Because of the thermal
-        variation time scales, this effect is usually most prominent in long observations (more
-        than about 50 ksec).
+        Thermal cycling on the spacecraft can result in an apparent
+        temporal drift of the sky position of an X-ray source during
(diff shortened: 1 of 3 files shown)


DougBurke:
This definitely needs testing by people other than me, and with different CIAO set ups.

kglotfelty:
Tagging @jeanconn who can hopefully review and sign off on change

jeanconn:
Also hitting @taldcroft .

ciao-4.11/contrib/share/doc/xml/correct_periscope_drift.xml
```diff
@@ -54,31 +54,35 @@ verbose=2
...
 ------------------------------------------------------------
 Writing out corrected aspect solution file to driftcorr_asol1.fits
-To review fit see correction plots in:
-corr_fit_yag.png
-corr_data_yag.png
-corr_fit_zag.png
-corr_data_zag.png
+        To review fit see correction plots in:
+                corr_fit_yag.png
+                corr_data_yag.png
```
jeanconn: I'm fine with not automatically popping up the plots when they are made; especially when the tool output still has this language about where they are. 

 We might want to add some stronger language right in the output (Do not use fit without reviewing plots... etc).

DougBurke:
The code now uses whatever plotting backend that Sherpa uses, rather than forcing matplotlib. The plots are now created by Sherpa itself, rather than manually, which means that they look slightly different than previous versions (but not, I claim, in any significant manner).

We still have the situation that the ChIPS plots appear briefly but the matplotlib ones do not.

DougBurke:
@jeanconn - I plan to rebase this a bit to clean up the commits, but are the documentation changes in  2383825 strong enough, or do we need more?

DougBurke:
Latest versions of the plots are:

# matplotlib

corr_data_zag
corr_fit_zag

# chips

corr_data_zag
corr_fit_zag

DougBurke:
@jeanconn - should both plots be in ks or is it worth keeping the raw data in s?

DougBurke:
I'm going to merge this and we can tweak things later if required.

jeanconn:
Sounds good regarding merge.  Regarding ks vs s, most likely the half original plots were in s because I forgot to change them to ks, not because of any real driver.  I don't feel strongly about this.

PR #203, opened 2019-02-12
Fix bug in save_marx_spectrum

### LINKED ISSUE #194: sherpa_contrib.marx fluxes
@hamogu Moritz: I was just testing out your sherpa_contrib.marx routines and I'm puzzled.  From what I understand, input spectra in MARX should have flux units of [photons/cm**2/s/keV] (https://space.mit.edu/cxc/marx/indetail/models.html#parameter-SpectrumFile) while ChaRT requires flux in [photons/cm**2/s].

Using the same absorbed power-law model and the sherpa_contrib.chart routines and sherpa_contrib.marx routines, the resulting spectrum appear to have the same fluxes, which I wouldn't have expected given the different unit requirements (see attached files).

[source_flux_chart.txt](https://github.com/cxcsds/ciao-contrib/files/2681188/source_flux_chart.txt)
[source_flux_marx.txt](https://github.com/cxcsds/ciao-contrib/files/2681189/source_flux_marx.txt)

Can you please confirm whether the documentation or module is problematic?

hamogu:
MARX expects the flux in per bin width (i.e. per keV), contrary to
ChaRT. Thus, data calculated with the ChaRT routines internally must
be divided by the bin width.

Tests have been added for a very simple model to at least verify this
behaviour.

Please try out the tests and verify that is passes, I simply don't know how
them best. (I can probably figure it out, but if it's easy for you, please try.)

fixes #194

### DIFF
--- a/ciao-4.11/contrib/Changes.CIAO_scripts
+++ b/ciao-4.11/contrib/Changes.CIAO_scripts
@@ -18,6 +18,13 @@ Updated scripts
     Corrects logic error when an invalid user query (ie invalid column
     names) triggers an unhelpful error message.
 
+Updated Python modules
+
+  sherpa_contrib.marx
+
+    The save_marx_spectrum function now normalizes the output by the bin
+    width, as expected by MARX.
+
 ## 4.11.1 - December 2018 ##
 
 Updated scripts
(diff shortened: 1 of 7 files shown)


hamogu:
OK, I made myself a local CIAO installation (usually I use out network installation) so that I could run the tests. And, of course, they failed, but should be fixed now.

DougBurke:
Thanks. The testing for this stuff is not as easy as it should be, for
historical reasons.

hamogu:
Anything that needs to be done or can this be merged?

DougBurke:
I updated the docs and merged. Ta.

PR #204, opened 2019-02-22
deflare converted to matplotlib

nplee:
updated deflare and lightcurves.py to use matplotlib instead of pychips.  Paramerter and ahelp files also updated.  Output plots saved now to PDF by default, but PS, EPS, and PNG formats are also supported.


DougBurke:
@nplee - I've just rebased your PR to reflect recent changes to the master branch and to remove all the merges you did.

DougBurke:
@nplee - and my latest push removes the unrelated specextract2 commits

PR #222, opened 2019-05-17
Dax mega update v5

kglotfelty:
This release includes major changes to dax affecting most of the individual analysis tasks.

    Output files:
    Users can now select the directory where they want the DAX output file to be written. There is a new dax.par parameter file where users can specify the default output directory, outdir. Alternatively, user can set the DAX_OUTDIR environment variable, or they can change the output directory via the new 'Change DAX output directory' analysis task.

    Statistics
    - Removed individual dmstat statistics
    - Reformatted dmstat output
    - Reformatted moments output
    - Removed centroid snap task from menu
    - Improved Help
    - Net counts are now computed for the selected regions (if any) or all regions if none are selected. Also now uses the image as currently displayed rather (eg smoothed, filtered).

    Histograms :
    All of the histogram plots are now created using ds9 built-in "BLT" plotting package. Users can use the options in the menu bar to save (print) the figure and adjust the plot style (linear|log, fonts, colors). All of the tasks also now save the output file in the DAX_OUTDIR (as discussed above).
    - New convert_ds9_region_to_ciao_stack script replaces old awk script to expand ds9 regions: panda, epanda, bpanda, mutli-radii annulus, multi-radii ellipse, mutli-radii box. Also handles excluded regions.
    - Radial profile task now uses image as displayed (eg smoothed/etc) instead of original file on disk.
    - User can now select pixels are arcsec for radial profile task
    - Extract along polygon now also plots data vs row-number, x, y, radius, or angle.
    - Improved Help

    Coordinates
    - Removed individual coordinates; now always display all coordinates.
    - Reformatted 'All' coordinates
    - Coordinates are now computed at: The location of the crosshair, if in crosshair mode; Or, the center of the 1st selected circle|box|ellipse|annulus; Or, the center of the 1st non-selected circle|box|ellipse|annulus.
    - Chandra Field-of-View task now run inside a wrapper script so that any error messages/etc are shown.
    - Improved Help

    Detect
    - All detect tasks are now run with data as displayed rather than going back to file on disk.
    - All detect tasks are now run inside a wrapper script so that any error messages/etc are shown.
    - celldetect is now run with fixedcell=12
    - Improved Help

    Region
    - All region tasks are now run with data as displayed rather than going back to file on disk.
    - All region tasks are now run inside a wrapper script so that any error messages/etc are shown.
    - New contour task runs dmcontour to create regions at specified levels. Interior holes are automatically excluded.
    - Improved error checking in PSF Size task
    - Improved Help

    Sherpa
    - Replaced chips plots with native ds9 blt plots.
    - Spectral fit now gives user the choice of statistic (chi2gehrels|chi2xspecvar|cash|cstat|wstat), minimization method (levmar|neldermead|moncar), absorption model(xswabs|xsphabs)
    - Removed clean cache task
    - 2D Image fit now saves output model, residuals, and fitting commands. Model and Residuals are now displayed in separate, tiled, frames.
    - New Radial Profile Fit task. Computes and fits radial profile with 1 or 2 additive model components from gauss1d|beta1d|const1d|lorentz1d|polynom1d|powlaw1d
    - Improved Help

    Image Processing
    - All tasks are now run inside a wrapper script so that any error messages/etc are shown.
    - Updated Source Fill task to no longer rely on user Grouping source and background regions. The script now requires 1-to-1 src-to-background region matches.

    Change DAX output directory
    New task to allow users to change DAX output directory during the current analysis session. The directory selected is not saved. Use pset dax outdir=directory to permanently change the output directory.


DougBurke:
Is the following caveat, from `dax.xml`, still valid with the introduction of `convert_ds9_region_to_ciao_stack`?

```
    <ADESC title="Known Limitations">
      <LIST>
        <ITEM>
	  The stacked shapes (panda, annulus, elliptical annulus, box
	  annulus, etc.) only work with the radial profile task.
	</ITEM>        
      </LIST>
   </ADESC>
```

PR #225, opened 2019-06-06
remove chips import from dax specfit script

kglotfelty:
While I removed `chips` plotting from the script, I missed the initial `import`.  

This does not affect ciao-4.11 since `chips` is still in the system; but does affect ciaox.


DougBurke:
I know I've just closed this, but I thought it was a sensible place to ask: should we also remove `bin/startchips` ?

kglotfelty:
Yup, it can be.  It doesn't hurt being there even in ciaox, but we should delete before 4.12

Issue #224, opened 2019-06-05
make_instmap_weights chips deprecation


kglotfelty:
The `make_instmap_weights` script imports `pychips` via `sherpa_contrib.utils` 

https://github.com/cxcsds/ciao-contrib/blob/70fcbdc11158a719bafa0bf607336e17925b9ba4/ciao-4.11/contrib/lib/python3.5/site-packages/sherpa_contrib/utils.py#L66

We'll need to update to use `matplotlib`

PR #227, opened 2019-06-06
Always set XSPEC settings in instmap weights

DougBurke:
Always set the XSPEC abundance and cross-section tables, since the defaults have changed for XSPEC 12.10.1 (from `bcmc` to `vern`). The parameter default is left at `bcmc` since it is quicker and the differences in the resulting exposure-map are expected to be small compared to other uncertainties.


DougBurke:
I'm moving the #224 work to a different PR. Once I write it.

PR #228, opened 2019-06-06
Update non-radial-profile sherpa plots to support matplotlib (fix #224)

### LINKED ISSUE #224: make_instmap_weights chips deprecation

DougBurke:
Fix the plotting in `sherpa_contrib` - for the `utils`, `chart`, and `marx` modules - so that it uses the Sherpa plotting objects (so is backend agnostic).

 - [x] `sherpa_contrib.utils.plot_instmap_weights`
    - adds the `clearwindow` argument
    - now plots the weights after downcasting to `float32` as there was "noise" when using `float64`
      which matplotlib was displaying
 - [x] `sherpa_contrib.chart.plot_chart_spectrum`
   - adds the `overplot` and `clearwindow` arguments
   - moves the model name to the plot title and drops the dataset label
 - [x] `sherpa_contrib.marx.plot_marx_spectrum`
   - adds the `overplot` and `clearwindow` arguments
   - moves the model name to the plot title and drops the dataset label
   - plots the Y axis in units of `photon/cm^2/s/keV` to match the data needed by MARX
 - [X] update the ahelp documentation
 - [X] add the changelog entry

# Note

This does not make all Sherpa plots use matplotlib, as the `sherpa_contrib.profiles.prof_XXX` routines still need to be updated, but that's for another PR.

### DIFF
--- a/ciao-4.11/contrib/lib/python3.5/site-packages/sherpa_contrib/chart.py
+++ b/ciao-4.11/contrib/lib/python3.5/site-packages/sherpa_contrib/chart.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2009, 2010, 2015, 2016
+# Copyright (C) 2009, 2010, 2015, 2016, 2019
 #           Smithsonian Astrophysical Observatory
 #
 #
@@ -78,8 +78,6 @@ def _get_chart_spectrum(id=None, elow=None, ehigh=None, ewidth=None,
             # b) Assume this means the dataset is not derived from the
             #    PHA class
             arf = None
-        except:
-            raise
 
         if arf is None:
             emsg = "No ARF found for dataset {} ".format(repr(id)) + \
@@ -234,7 +232,7 @@ def save_chart_spectrum(outfile, clobber=True, verbose=True,
                   format=format, clobber=clobber)
 
     if verbose:
-        print ("Created: {}".format(outfile))
+        print("Created: {}".format(outfile))
 
 
 def plot_chart_spectrum(**kwargs):
@@ -270,7 +268,7 @@ def plot_chart_spectrum(**kwargs):
                     "Dataset: {}".format(repr(vals["id"])),
                     ["halign", 1])
 
-    except:
+    except Exception:
         c.discard_undo_buffer()
         raise
 
--- a/ciao-4.11/contrib/lib/python3.5/site-packages/sherpa_contrib/chart.py
+++ b/ciao-4.11/contrib/lib/python3.5/site-packages/sherpa_contrib/chart.py
@@ -131,9 +131,10 @@ def get_chart_spectrum(id=None, elow=None, ehigh=None, ewidth=None,
         dataset must have a grid and source model defined.
     elow, ehigh, ewidth : number or None
         Only used if all three are given, otherwise the grid
-        of the data set is used. The elo value is the start
-        of the low-energy bin, ehigh is the end of the high-energy
-        bin, and ewidth the bin width, all in keV.
+        of the data set is used. The elo value gives the start
+        of the grid (the left edge of the first bin) and ehi
+        is the end of the grid (the right edge of the last bin),
+        and ewidth the bin width, all in keV.
     norm : number or None
         Multiply the flux values by this value, if set.
 
@@ -206,9 +207,10 @@ def save_chart_spectrum(outfile, clobber=True, verbose=True,
         dataset must have a grid and source model defined.
     elow, ehigh, ewidth : number or None
         Only used if all three are given, otherwise the grid
-        of the data set is used. The elo value is the start
-        of the low-energy bin, ehigh is the end of the high-energy
-        bin, and ewidth the bin width, all in keV.
+        of the data set is used. The elo value gives the start
+        of the grid (the left edge of the first bin) and ehi
+        is the end of the grid (the right edge of the last bin),
+        and ewidth the bin width, all in keV.
     norm : number or None
         Multiply the flux values by this value, if set.
 
@@ -256,9 +258,10 @@ def plot_chart_spectrum(id=None, elow=None, ehigh=None, ewidth=None,
         dataset must have a grid and source model defined.
     elow, ehigh, ewidth : number or None
         Only used if all three are given, otherwise the grid
-        of the data set is used. The elo value is the start
-        of the low-energy bin, ehigh is the end of the high-energy
-        bin, and ewidth the bin width, all in keV.
+        of the data set is used. The elo value gives the start
+        of the grid (the left edge of the first bin) and ehi
+        is the end of the grid (the right edge of the last bin),
+        and ewidth the bin width, all in keV.
     norm : number or None
         Multiply the flux values by this value, if set.
     overplot : bool, optional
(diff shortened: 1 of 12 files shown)


DougBurke:
@hamogu I've updated the marx plotting code so that it is now backend agnostic (uses the sherpa choice). As part of this I've tweaked the module slightly as discussed in @c221ee7 - do these changes look okay (documentation and the change of `plot_marx_spectrum` to plot `photon/cm^2/s/keV` rather than `photon/cm^2/s`, as I wanted these routines to plot the data written out to the file [there is an argument that I'm plotting a histogram and perhaps it should be a curve to match the file contents]

ciao-4.11/contrib/lib/python3.5/site-packages/sherpa_contrib/chart.py
```diff
@@ -205,6 +201,16 @@ def save_chart_spectrum(outfile, clobber=True, verbose=True,
         Should a message be displayed indicating that the file
         has been created? This is only used when outfile is not
         ``None``.
+    id : int or string or None
+        If id is None then the default Sherpa id is used. This
+        dataset must have a grid and source model defined.
+    elow, ehigh, ewidth : number or None
+        Only used if all three are given, otherwise the grid
+        of the data set is used. The elo value is the start
+        of the low-energy bin, ehigh is the end of the high-energy
```
hamogu: low-energybin -> bin of lowest energy and the same for high-energy-bin. When we talk about Chandra we say things like "The low-energy signal", whereas here "low-energy" cna actually be a high energy, e.h. `elow=9, ehigh=10, ewdith=0.1`, so I suggest rewording to avoid confusion.

ciao-4.11/contrib/lib/python3.5/site-packages/sherpa_contrib/chart.py
```diff
@@ -228,50 +234,74 @@ def save_chart_spectrum(outfile, clobber=True, verbose=True,
...
+    Parameters
+    ----------
+    id : int or string or None
+        If id is None then the default Sherpa id is used. This
+        dataset must have a grid and source model defined.
+    elow, ehigh, ewidth : number or None
+        Only used if all three are given, otherwise the grid
+        of the data set is used. The elo value is the start
+        of the low-energy bin, ehigh is the end of the high-energy
+        bin, and ewidth the bin width, all in keV.
```
hamogu: see comment above

ciao-4.11/contrib/lib/python3.5/site-packages/sherpa_contrib/chart.py
```diff
@@ -228,50 +234,74 @@ def save_chart_spectrum(outfile, clobber=True, verbose=True,
...
+    hplot.xlo = vals['xlo']
+    hplot.xhi = vals['xhi']
+    hplot.y = vals['y']  # .astype(np.float32)
+    hplot.xlabel = 'Energy (keV)'
+    hplot.title = 'ChaRT Spectrum: {}'.format(vals['model'])
+
+    # LaTeX support depends on the backend
+    if backend.name == 'pylab':
+        hplot.ylabel = 'Flux (photon cm$^{-2}$ s$^{-1}$)'
+    elif backend.name == 'chips':
```
hamogu: Are we going to release this while chips is still alive?

ciao-4.11/contrib/lib/python3.5/site-packages/sherpa_contrib/marx.py
```diff
@@ -129,6 +127,16 @@ def save_marx_spectrum(outfile, clobber=True, verbose=True,
...
         has been created? This is only used when outfile is not
         ``None``.
+    id : int or string or None
+        If id is None then the default Sherpa id is used. This
+        dataset must have a grid and source model defined.
+    elow, ehigh, ewidth : number or None
+        Only used if all three are given, otherwise the grid
+        of the data set is used. The elo value is the start
+        of the low-energy bin, ehigh is the end of the high-energy
+        bin, and ewidth the bin width, all in keV.
```
hamogu: see comment above.

ciao-4.11/contrib/lib/python3.5/site-packages/sherpa_contrib/marx.py
```diff
@@ -68,7 +68,7 @@ def get_marx_spectrum(id=None, elow=None, ehigh=None, ewidth=None,
     Returns
     -------
     (energ_hi, flux_bins)
-       The energies are in keV and fluxes in photons/cm^2/s.
+       The energies are in keV and fluxes in photons/cm^2/s/keV.
```
hamogu: yes!

ciao-4.11/contrib/lib/python3.5/site-packages/sherpa_contrib/marx.py
```diff
@@ -139,21 +147,64 @@ def save_marx_spectrum(outfile, clobber=True, verbose=True,
...
+    Parameters
+    ----------
+    id : int or string or None
+        If id is None then the default Sherpa id is used. This
+        dataset must have a grid and source model defined.
+    elow, ehigh, ewidth : number or None
+        Only used if all three are given, otherwise the grid
+        of the data set is used. The elo value is the start
+        of the low-energy bin, ehigh is the end of the high-energy
+        bin, and ewidth the bin width, all in keV.
```
hamogu: see above.

DougBurke:
Thanks for the comments @hamogu - I'll look at them tomorrow.

ciao-4.11/contrib/lib/python3.5/site-packages/sherpa_contrib/chart.py
DougBurke: I'm changing this to the following (in `chart.py` and `marx.py`). I am not 100% happy with using left/right edge, but I also don't think it is worth obsessing over too much.

```
    elow, ehigh, ewidth : number or None
        Only used if all three are given, otherwise the grid
        of the data set is used. The elo value gives the start
        of the grid (the left edge of the first bin) and ehi
        is the end of the grid (the right edge of the last bin),
        and ewidth the bin width, all in keV.
```

Perhaps I should centralize this documentation rather than copy-pasting, but that has a user cost (having to look it up), so I am not going to do that now.

ciao-4.11/contrib/lib/python3.5/site-packages/sherpa_contrib/chart.py
DougBurke: I think we should, and as it is a small code change to support I think it's worth it even if we don't (plus if I ever do decide to right a plotly / altair / ... backend for notebook use it reminds me to update this).

DougBurke:
Thanks for the review @hamogu - I'm going to assume I've addressed your comments sufficiently, but please holler if not :-)

hamogu:
Looks good to meOn Jun 7, 2019 8:54 AM, Doug Burke <notifications@github.com> wrote:Thanks for the review @hamogu - I'm going to assume I've addressed your comments sufficiently, but please holler if not :-)

PR #230, opened 2019-06-07
Remove DAX chips-startup code (fix #229)

### LINKED ISSUE #229: Remove start chips
Yup, it can be.  It doesn't hurt being there even in ciaox, but we should delete before 4.12

_Originally posted by @kglotfelty in https://github.com/cxcsds/ciao-contrib/pull/225#issuecomment-499499947_

DougBurke:
This code is not used as plotting in DAX is now handled by DS9 rather
than ChIPS. This is a follow on to the cleanup in #225 and fixes #229


DougBurke:
@kglotfelty I assume removing the `.tk` file as well is appropriate?

kglotfelty:
the `.tk` is still needed.  Not for `chips` , but it has the proc needed to change the dax output directory.

(We could rename it, but then that needs to be coordinated with CIAO wrapper)

DougBurke:
And the wrapper is - I assume - provided as part of the CIAO installation (hence the coordination)?

kglotfelty:
correct.

DougBurke:
So the `.tk` file has an `exec startchips` in it. Is this just going to fail quietly and not be noticed by the user? Or should the chips-related lines in it be removed as well? Which looks like it's more something you should do than me :-)

kglotfelty:
The latest version of that file has all that stuff commented out.  It could be cleaned up but should be ok ?

DougBurke:
Ahh, I wasn't sure of the syntax of the file, so wasn't sure if it was commented out or not. I'd prefer to remove the comments (in part because it makes grepping through the code base for 'chips' a bit less noisy), but it's not a big deal.

DougBurke:
So, something like the new version I've just pushed. I have not actually tried this to see if I've broken anything when changing `chips_startup.tk`

kglotfelty:
Yup, that works (copied into local build w/o any problems).

:shipit:

Issue #150, opened 2018-10-29
CIAO 4.11 remove specialized history propagation


kglotfelty:
With the history mechanism in crates turned on in CIAO 4.11, we should remove any specialized code we have that's propagating history using `cxcdm`, eg in @nplee 's `ecf_calc` 

https://github.com/cxcsds/ciao-contrib/blob/0ecfe7743c3a5e0e7e8b2339a69cbc3c2975eea2/ciao-4.10/contrib/bin/ecf_calc#L66

I thought that Pete's tools (`detilt`, `dewiggle`, `symmertize`) were also doing something like this but I'm not seeing it right now.

nplee:
addressed in latest revision of #235

DougBurke:
I'm closing this now that #235 has fixed this for `ecf_calc`. It can be re-opened if there are other scripts which need to be updated.

PR #235, opened 2019-06-07
ecf_calc plotting updated to matplotlib

nplee:
replace pychips with matplotlib in ecf_calc for plotting.

### DIFF
--- a/ciao-4.11/contrib/bin/ecf_calc
+++ b/ciao-4.11/contrib/bin/ecf_calc
@@ -20,11 +20,9 @@ Usage:
 
 __version__ = "CIAO 4.11"
 toolname = "ecf_calc"
-__revision__ = "07 June 2019"
+__revision__ = "12 June 2019"
 
 import numpy, tempfile, sys, os, paramio, cxcdm
-
-import six
 
 import pycrates as pcr
 
@@ -133,7 +131,7 @@ def radii_set(radius,bin):
 
 
 def pause():
-    six.moves.input("Press ENTER to close the ChIPS window and exit:")
+    input("Press ENTER to close the ChIPS window and exit:")
 
 
 def get_par(argv):
@@ -372,8 +370,7 @@ def plot_ecf(rad_file):
 
     matplotlib.use("TkAgg")
     matplotlib.interactive(True)
-    matplotlib.rcParams["toolbar"] = "None"    
-
+    
     try:
         detector = get_detnam(rad_file)
     except KeyError:
--- a/ciao-4.11/contrib/bin/ecf_calc
+++ b/ciao-4.11/contrib/bin/ecf_calc
@@ -18,9 +18,9 @@ Usage:
 
 """
 
-__version__ = "CIAO 4.9"
+__version__ = "CIAO 4.11"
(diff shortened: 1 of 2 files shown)


ciao-4.11/contrib/bin/ecf_calc
```diff
@@ -365,10 +364,15 @@ def get_ecf(infile,reg,outfile,x,y,radius,binsize,fraction,clobber):
...
 def plot_ecf(rad_file):
 
-    import pychips
+    import matplotlib
+    from matplotlib import pyplot as plt
+    from matplotlib.ticker import ScalarFormatter
+
+    matplotlib.use("TkAgg")
+    matplotlib.interactive(True)
+    matplotlib.rcParams["toolbar"] = "None"
```
DougBurke: I'm too lazy to RTFD, but is this script intended to just create an output file, or display the data for the user to see before exiting the script.

nplee: It's just a plot to the screen, there's no output to file.

DougBurke:
The ahelp file will need updating (to say Matplotlib rather than ChIPS and to add a "changes in this version ..." section).

ciao-4.11/contrib/bin/ecf_calc
DougBurke: Is there a reason to turn the toolbar off? Users might want to be able to save a png or zoom around in the figure (or am I over-thinking it)?

DougBurke:
Along with the changes in #150 I'd like to see the `import six` line removed - it is only used for `six.moves.input` which can be replaced with `input`.

nplee:
@DougBurke: I've updated the ahelp, replaced six.moves.input with input, and enabled the matplotlib toolbar.

ciao-4.11/contrib/share/doc/xml/ecf_calc.xml
```diff
@@ -191,17 +191,17 @@
...
-<!--//    <ADESC title="Changes in the December 2012 release">
+    <ADESC title="Changes in the December 2012 release">
       <PARA>
 	The script is new for in CIAO 4.5.
       </PARA>
-    </ADESC> //-->
+    </ADESC>
 
-    <ADESC title="Changes in the scripts 4.9.3 (May 2017) release">
+    <ADESC title="Changes in the June 2019 release">
```
DougBurke: @nplee - ta for this.

Can you move this before the Dec 2012 one (so we get the most-recent change first), and change the 2019 title to "Changes in the 4.11.4 (2019) release" [don't change the 2012 one]? I'm not sure what month this is going out in so am leaving it as `(2019)` for now.

DougBurke:
@nplee - do you think you can put the #150 changes in this pr or would you rather do them separately?

DougBurke:
I'm wondering if we still need the use of `input`, since we could just rely on the user closing the matplotlib window to indicate that they have finished. This would probably require removing setting the `interactive` setting for mpl.

You'd still probably need a screen message to indicate that the script was waiting. I also wonder if it would be annoying (since you have to use the mouse to close the window rather than just pressing return).

PR #240, opened 2019-06-11
Cleanup chips influences in the lightcurve code

DougBurke:
- remove chips references from the lightcurve code base and documentation
- fix a minor issue whereby underscores in labels would appear as '\_' in the plot (not sure if this is likely to happen, but it's a holdover from the ChIPS support)
- note that the erase argument is no-longer used, but don't remove it in case someone has code that sets it
- similarly the protect argument is no-longer used
- remove use of 'import six'
- minor style changes, as suggested by `pycodestyle`

### DIFF
--- a/ciao-4.11/contrib/bin/deflare
+++ b/ciao-4.11/contrib/bin/deflare
@@ -1,7 +1,8 @@
 #!/usr/bin/env python
 
 #
-#  Copyright (C) 2014, 2015, 2016, 2017, 2018, 2019  Smithsonian Astrophysical Observatory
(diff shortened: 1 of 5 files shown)


DougBurke:
@nplee - can you just look this over to make sure it is okay?

DougBurke:
The PR has been rebased to reflect the latest master branch

nplee:
# @DougBurke changes look good to me (or at least I don't see anything problematic)

DougBurke:
Ta

PR #241, opened 2019-06-11
Remove use of the six module

DougBurke:
This removes the use of the six module from most of the support code. It doesn't change the ChIPS code (`chips_contrib.xxx`) as this is going to be removed for CIAO 4.12, and leaves out some files as they are the subject of other PRs (lightcurves in #240, ecf_calc is in #235)

There has been some linting changes (addition and removal of spaces in most cases, but explicit error catching is the other major change) based on the report from `pycodestyle` (not all warnings and errors have been implemented).

### DIFF
--- a/ciao-4.11/contrib/bin/download_obsid_caldb
+++ b/ciao-4.11/contrib/bin/download_obsid_caldb
@@ -29,7 +29,7 @@ import ciao_contrib.logger_wrapper as lw
 
 
 toolname = "download_obsid_caldb"
-__revision__ = "11 June 2019"
+__revision__ = "18 June 2019"
 
 lw.initialize_logger(toolname)
 
@@ -487,18 +487,23 @@ def get_all_products():
     with open(cidx, "r+") as fp:
         lines = fp.readlines()
 
-    chandra = [x for x in lines if x.startswith("CHANDRA")]
-    ciao = [x for x in chandra if x.split()[1] not in ['EPHIN', 'PCAD', 'SIM']]
-    ciao = [x for x in ciao if 'pimms' not in x.lower()]
-    ciao = [x.replace(" CALDB ", " {} ".format(cdir)) for x in ciao]
-    idxfiles = [os.path.join(x.split()[2],
-                             os.path.join(x.split()[3], x.split()[4]))
-                for x in ciao]
+    idxfiles = []
+    for line in lines:
+        if not line.startswith('CHANDRA'):
+            continue
+        if line.split()[1] in ['EPHIN', 'PCAD', 'SIM']:
+            continue
+        if 'pimms' in line.lower():
+            continue
+
+        paths = line.replace(' CALDB ', ' {} '.format(cdir)).split()
+        path = os.path.join(paths[2], paths[3], paths[4])
+        idxfiles.append(path)
 
     retval = {}
     for idx in idxfiles:
         # Quality of 5 is bad, not retrieved
-        tab = read_file(idx+"[cal_qual=:4.9]")
+        tab = read_file(idx + "[cal_qual=:4.9]")
         # get list of unique code-names
         _prods = list(set(tab.get_column("cal_cnam").values))
         try:
--- a/ciao-4.11/contrib/bin/download_obsid_caldb
+++ b/ciao-4.11/contrib/bin/download_obsid_caldb
@@ -68,7 +68,7 @@ class LoggingFile:
             self.bytes = 0
         else:
             if startfrom > filesize:
-                raise ValueError("startfrom > filesize ({} vs {})".format(startfrom,filesize))
+                raise ValueError("startfrom > filesize ({} vs {})".format(startfrom, filesize))
             self.bytes = startfrom - 1
         self.hashes = 0
         self.filesize = filesize
@@ -102,21 +102,21 @@ class FtpFile():
     """
     Little wrapper class to get the file size
     """
-    def __init__( self, ftp, path ):
+    def __init__(self, ftp, path):
         self.ftp = ftp
         self.path = path
 
-    def get_ftp_filesize( self):
+    def get_ftp_filesize(self):
 
         self.filesize = None
 
-        def parse_size( line ):
+        def parse_size(line):
             toks = line.split()
             self.filesize = int(toks[4])
 
         self.ftp.retrlines("LIST " + self.path, parse_size)
 
-        return( self.filesize )
+        return self.filesize
 
 
 def parse_url():
@@ -164,7 +164,7 @@ def parse_url():
     user = "anonymous" if not url.username else url.username
     pswd = "anonymous@ciao_contrib."+toolname if not url.password else url.password
 
-    verb2("Using FTP host='{}' dir='{}' user='{}' password='{}'".format( host, path, user, pswd))
+    verb2("Using FTP host='{}' dir='{}' user='{}' password='{}'".format(host, path, user, pswd))
     return host, path, user, pswd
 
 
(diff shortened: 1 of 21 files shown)


DougBurke:
@kglotfelty I've pinged you so you can check the modules you feel any affinity or ownership for, but feel free to check more. I'm assuming that these changes are going to be "verified" by a regression-test run.

kglotfelty:
Just a quick note ... I'm not sure relying on the regression tests is sufficient here.  Our coverage is mostly limited to stuff that's run via a higher level command-line tool.  We have minimal coverage of stuff that's run at the python level.

I'm thinking things like the proptools; we may have 1 specific usage in specextract but not wide coverage of the options.  Similarly will be the sherpa rprof stuff (I'm not sure we have any coverage in our suite.)

ciao-4.11/contrib/lib/python3.5/site-packages/ciao_contrib/cda/data.py
```diff
@@ -199,16 +200,20 @@ def extract_file_type(fname):
...
         return None
 
-    if fn == "oif.fits": return "oif"
-    if fn.lower() == "00readme": return "readme"
-    if fn.find("vv") != -1: return "vv"
+    if fn == "oif.fits":
+        return "oif"
+    if fn.lower() == "00readme":
+        return "readme"
+    if fn.find("vv") != -1:
```
kglotfelty: Unrelated to this PR!

Since we don't care where it's at in the string should we maybe instead use

```python
if 'vv' in fn:
```

instead? (same below in the `known_file_types` loop).

ciao-4.11/contrib/lib/python3.5/site-packages/ciao_contrib/cda/data.py
```diff
@@ -334,14 +340,16 @@ def is_type(self, types):
...
         for ft in types:
-            if ft == self.filetype: return True
+            if ft == self.filetype:
+                return True
         return False
 
     def is_format(self, formats):
         """Given a list of file formats, returns True if
         the file is one of these formats."""
         for ff in formats:
```
kglotfelty: Also unrelated to this PR:

Could this just be

```python
    return ff in self.fileformat
```
?

ciao-4.11/contrib/lib/python3.5/site-packages/ciao_contrib/cda/csccli.py
```diff
@@ -818,24 +813,23 @@ def fix_dm_no_g_type( page ):
...
+      to be "(Fnumber)"
     """
     import re as re
-    
-    fixed = re.sub( '\(G[0-9\.]*\)',
-                lambda x: x.group(0).replace("G", "F"),
-                page )
-    return fixed
 
+    fixed = re.sub(r'\(G[0-9\.]*\)',
```
kglotfelty: Needs review.

Since string values are already escaped, the `\` now gets into the regular expression which may foul up the parsing.

ciao-4.11/contrib/bin/download_obsid_caldb
```diff
@@ -476,64 +472,66 @@ def get_all_products():
...
-    
+
     chandra = [x for x in lines if x.startswith("CHANDRA")]
-    ciao = [x for x in chandra if x.split()[1] not in  ['EPHIN', 'PCAD', 'SIM' ]]
+    ciao = [x for x in chandra if x.split()[1] not in ['EPHIN', 'PCAD', 'SIM']]
     ciao = [x for x in ciao if 'pimms' not in x.lower()]
-    ciao = [ x.replace(" CALDB ", " {} ".format(cdir)) for x in ciao]
-    idxfiles = [ os.path.join( x.split()[2], os.path.join( x.split()[3],x.split()[4] )) for x in ciao ]
+    ciao = [x.replace(" CALDB ", " {} ".format(cdir)) for x in ciao]
+    idxfiles = [os.path.join(x.split()[2],
```
kglotfelty: Not related to this PR.

This can be simplified ; `os.path.join` takes multiple dirs.

ciao-4.11/contrib/bin/download_obsid_caldb
```diff
@@ -467,6 +461,8 @@ def retrieve_files_by_tool( ftp, rootdir, tool, infile, outdir, clobber, missing
 
         products = cc.search
         retrieve_caldb_file( ftp, rootdir, products, outdir, clobber, missing )
+
+        # What does the following line do?
```
kglotfelty: At the C/C++ level it free memory/etc.  Presumably something similar in python.  

[The whole caldb module is not very pythonic.]

ciao-4.11/contrib/bin/download_obsid_caldb
```diff
@@ -752,47 +747,43 @@ def main():
...
 
-    infiles= pars["infile"]   
-    outdir = pars["outdir"]   
-    missing = ( "yes" == pars["missing"] )
-    background_files = ("yes" == pars["background"] ) 
-    clobber = ("yes" == pars["clobber"]) 
-        
+    infiles= pars["infile"]
+    outdir = pars["outdir"]
+    missing = "yes" == pars["missing"]
```
kglotfelty: I think the `()` are needed here if nothing more than just for readability.

Can maybe use `is` instead of `==` to make it more clear

kglotfelty: Nope, we cannot use `is` here.

The syntax works here so it's _ok_, but I think with the `()` the code is much more readable.

ciao-4.11/contrib/lib/python3.5/site-packages/ciao_contrib/cda/csccli.py
kglotfelty: This works ok here.

There are other similar changes in the `coords` package ... definitely sure we're not hitting all those code paths in the regression tests so not sure if there will be an change in behavior.

ciao-4.11/contrib/lib/python3.5/site-packages/ciao_contrib/cda/data.py
DougBurke: It's been so long since I wrote this (I think it's mine) that I'll have to review the code to see what on earth it's doing!

ciao-4.11/contrib/lib/python3.5/site-packages/ciao_contrib/cda/csccli.py
DougBurke: If you don't add a leading r to the string then tools like `pycodestyle` complain. Whether that's a good or bad thing I don't know.

ciao-4.11/contrib/bin/download_obsid_caldb
DougBurke: But if it's a function then shouldn't it be `cc.close()`? Then again you might say that about `cc.search` so all bets are off here, as you say.

From my limited testing the `close` attribute seemed to be `None` (if memory serves).

kglotfelty:
exactly ... it's "I just learned about `@properties` decorator so let's use
them even if it doesn't make any sense".

ciao-4.11/contrib/lib/python3.5/site-packages/ciao_contrib/cda/data.py
DougBurke: Done

ciao-4.11/contrib/lib/python3.5/site-packages/ciao_contrib/cda/data.py
DougBurke: Done

ciao-4.11/contrib/bin/download_obsid_caldb
DougBurke: I re-wrote this to be a single loop, rather than a set of list comprehensions. I think it's easier to see the logic of what is/isn't included. Assuming I haven't botched something up (as the logic checks are now inverted).

ciao-4.11/contrib/bin/download_obsid_caldb
DougBurke: Replaced my comment with another comment for when I next look at this...

ciao-4.11/contrib/bin/download_obsid_caldb
DougBurke: I ended up moving the check into a separate function.

PR #248, opened 2019-06-25
chandra_repro: support HRC obs w/ 0 ontime

### LINKED ISSUE #239: chandra_repro HRC L2 ONTIME=0
Obsid 1206 has 0 good-time.  It is HRC and failing at the `hrc_dtfstats` step:

```
 An error occurred while running 'chandra_repro':
  
  Processing input directory '/Volumes/Chandra/Archive/1206'
  
  
  Running hrc_build_badpix to create new bad pixel file...
  
  Running hrc_process_events to reprocess the evt1.fits file...
  Filtering the evt1.fits file on grade and status...
  Applying the good time intervals from the flt1.fits file...
  
  Running hrc_dtfstats to recompute the average dead time corrections...
  # chandra_repro (19 November 2018): ERROR An error occurred while running 'hrc_dtfstats':
    #   (CIAO 4.11): ERROR: Could not open file '/Volumes/Chandra/Archive/1206/repro/hrcf01206_repro_evt2.fits[GTI]'.
```
Need to handle this case a bit more gracefully (probably copy L1 DTF value w/ warning)

kglotfelty:
This fixed #239 .  With 0 ontime, the L2 GTIs are missing (DM omits them) so instead we use the flt1 file.  It's still 0|NULL but it exists which is what was causing the problem.  Elsewhere in the script it issues a warning about 0 ONTIME so no need to add on here.

### DIFF
--- a/ciao-4.11/contrib/bin/chandra_repro
+++ b/ciao-4.11/contrib/bin/chandra_repro
@@ -34,7 +34,7 @@ Aim:
 """
 
 toolname = "chandra_repro"
-version = "13 May 2019"
+version = "11 Jun 2019"
 
 # import standard python modules as required
 #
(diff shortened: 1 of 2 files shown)


DougBurke:
Can an entry be added to the @contrib/CHANGELOG.whatever@ file?

kglotfelty:
changelog updated in separate pull request, 10f5e7e4a221d8f

Issue #236, opened 2019-06-10
coords.chandra _PNT keywords


DougBurke:
I just tried `coords.chandra.sky_to_chandra` on a l3 file (I'm updating the ahelp file to remove chips) and noticed that it failed

```
chips In [5]: sky_to_chandra(keywords, 4096.5, 4096.5)
KeyError: 'RA_PNT'
```

/pool14/dburke/sens3.fits is an example

Is this because the routines arent' designed for L3 products (e.g. merged), or because the module hasn't been updated for recent-ish changes, or some other nefarious reason?

DougBurke:
I've edited the title since I see the same thing with a L2 product

kglotfelty:
which L2 file?  

The PNTs should be there except for the multi-obi obsids.

DougBurke:
Sorry, I was imprecise. By L2 I meant a file created from L2. It was actually an output from `flux_obs` but I now can't remember if it was a merged file or from a single obsid.

kglotfelty:
Merged products are invalid since the default merging rules are to omit those keys.  I also see SIM_* and D*_AVG are missing.

DougBurke:
It was to a merged file.

DougBurke:
Is it easy to come up with a list of required keywords?

kglotfelty:
It's in one of the internal doc strings but not a particularly useful place for end-users

```python
    Input is a dictionary of keyword values that must contain
    the following values:
        
        INSTRUME
        DETNAM
        RA_NOM
        DEC_NOM
        RA_PNT
        DEC_PNT
        ROLL_PNT
        SIM_X
        SIM_Y
        SIM_Z
        DY_AVG
        DZ_AVG
        DTH_AVG
```

DougBurke:
I'm closing this now that #249 has been merged.

PR #249, opened 2019-06-25
coords.chandra added error checking

kglotfelty:
This adds some error checking to catch case like #236 where keywords missing, most like due to merged products.

I'm sure there could be more error checking done, but this is a lot better than before.


ciao-4.11/contrib/lib/python3.5/site-packages/coords/chandra.py
```diff
@@ -44,6 +48,9 @@ def _setup_sim( keyword_list ):
     if all( [x in keyword_list for x in ["SIM_X", "SIM_Y", "SIM_Z"] ]):
         sim = [ keyword_list["SIM_X"], keyword_list["SIM_Y"], keyword_list["SIM_Z"] ]
     else:
+        logger.verbose0("WARNING: Missing one or more of the following keywords: "+
```
DougBurke: This could just use the @v0@ symbol you created above :-)

DougBurke:
Can an entry be added to the changelog for this, mainly for advertising purposes?

kglotfelty:
changelog updated in  10f5e7e4a221d8f

DougBurke:
I rebased this to avoid the merges.

PR #118, opened 2018-03-30
WIP: Support XSPEC 12.9.1 (fix #99)

DougBurke:
This adds support for the convolution models added in XSPEC 12.9.1:

   clumin, cpflux, rfxconv, rgsxsrc, vashift, vmshift, xilconv

Notes
------

Limited testing. Need to better document `xilconv` support, since the data file will need to be downloaded from "somewhere". As this requires the as-yet-undocumented/worked out support for extending energy grids it is a lower priority.

Needs to be re-done to support XSPEC 12.10.0 convolution models too.

PR #254, opened 2019-07-22
WIP: Update convert_xspec_user_model for CIAO 4.11

DougBurke:
This is a rebase of #118 but done as a separate PR since it was quite a messy rebase. The original message follows.

----

This adds support for the convolution models added in XSPEC 12.9.1:

clumin, cpflux, rfxconv, rgsxsrc, vashift, vmshift, xilconv

Notes
Limited testing. Need to better document xilconv support, since the data file will need to be downloaded from "somewhere". As this requires the as-yet-undocumented/worked out support for extending energy grids it is a lower priority.

Needs to be re-done to support XSPEC 12.10.0 convolution models too.


DougBurke:
Rebased to remove a binary file that was accidentally committed as part of 5ed8cf9

DougBurke:
Replaced by #431

PR #256, opened 2019-07-30
Work around bug in conda tcsh environment

kglotfelty:
conda messes up the PATH environment variable wreaking havoc on
setup scripts.  Use $path instead.


DougBurke:
I assume we can take this now (rather than waiting to go out with it with CIAO 4.12)?

kglotfelty:
Yes, it is independent of 4.12.

DougBurke:
I don't think this needs an entry in the changelog (although the ahelp file will need updating when we next release).

PR #260, opened 2019-08-06
Move to setup py

kglotfelty:
This is an initial attempt to provide a build script for the ciao-contrib tree.

I had to flatten the `lib` directories to get `setup.py` to install directly into the `site-packages` directory; the other `bin` and `data` dirs were okay to install from where they have lived.  This probably needs review/rework.

```bash
conda create -n junk python=3.5
conda activate junk
python setup.py install
 - or -
python setup.py install --prefix=$CONDA_PREFIX/contrib
```
Note: the shebang's in the bin/ dir are automatically replaced with the hard-coded python path during the install.


DougBurke:
What do you think about having a top-level `src/` directory in which all the files that will be installed are to go? The idea is to make it a bit easier to separate out "scaffolding" (e.g. `mk_runtool.py`) from files we want to install.

DougBurke:
Here's the egg (from a python 3.7 conda install), so we need to add some metadata to `setup.py`

```
% cat lib/python3.7/site-packages/ciao_contrib-4.12.0-py3.7.egg-info 
Metadata-Version: 1.0
Name: ciao-contrib
Version: 4.12.0
Summary: CIAO Contributed scripts
Home-page: http://github.com/cxcsds/ciao-contrib/
Author: CXCSDS and Friends
Author-email: cxchelp@cfa.harvard.edu
License: UNKNOWN
Description: UNKNOWN
Platform: UNKNOWN
```

DougBurke:
So, I think the path for "classic" CIAO would be for `mk_script_tarfile` to do something like

```
python setup.py install --prefix=<some-temp-directory>/ciao-4.12/contrib
```

and then tar up the `ciao-4.12/` directory. Does this sound about right?

DougBurke:
To look at: `python setup.py install` reports for me

```
running install
running build
running build_py
package init file 'sherpa_contrib/tests/__init__.py' not found (or not a regular file)
package init file 'sherpa_contrib/xspec/tests/__init__.py' not found (or not a regular file)
package init file 'sherpa_contrib/tests/__init__.py' not found (or not a regular file)
package init file 'sherpa_contrib/xspec/tests/__init__.py' not found (or not a regular file)
running build_scripts
running install_lib
...
```

which is a reminder that we should add testing to this, so we can say `python setup.py test`, as well as to look at whether we need to add `__init__.py` files.

DougBurke:
Since I can never find this when I need it: https://packaging.python.org/guides/distributing-packages-using-setuptools/#distributing-packages

kglotfelty:
w.r.t. a toplevel `src` directory ...  It's easy to move everything except the modules into a `src` directory.

The problem is that the modules install into a directory using the same directory name -- so `coords` would install into `pythonX.Y/site-packages/src/coords` .  I've tried to use the `package_dirs` option but couldn't get it to work.

kglotfelty:
w.r.t.  `mk_script_tarfile` -- Yes.

kglotfelty:
Actually for `mk_script_tarfile` it is more complicated.

`setup.py` replaces the shebang with the `prefix/bin` location.  If we want to keep `/usr/bin/env python` then we'll have to play some different games.

kglotfelty:
Okay -- looks like we can pass in the executable argument 

```bash
python setup.py build --executable '/usr/bin/env python'  install --prefix=<dir>/ciao-4.12/contrib
```

DougBurke:
I'm okay to merge this, or do you think you need to do something more first @kglotfelty ?

kglotfelty:
I just make a quick tweak to setup.py to install the Changes and Readme's in the root dir.

Should be good to do.  Not perfect -- but a step in right direction.

PR #265, opened 2019-09-05
added catch on ValueError 

nplee:
a change to cr.delete_column throws a ValueError in CIAOX 09092019 build rather than AttributeError like in prior versions.


DougBurke:
Does not need a changelog and can go in before CIAO 4.12

Issue #262, opened 2019-08-20
merge_obs PSF map and header merging rules


nplee:
This came in from a recent helpdesk ticket (022153).

The user is trying to combined all the observations of 3C 9 (ObsIDs 1595, 17088, 18700, and 18701) with 'merge_obs' using both the 'maxsize' and 'psfmerge=exptime' arguments.

Two issues have come up:

1.) When 'maxsize' is set to some value other than INDEF, it supersedes both the 'binsize' and 'xygrid' parameters, so strange binning factors are used, which generate exposure maps that the pixel mask filter isn't happy with while trying to generate the PSF maps.  If 'maxsize' is being used, then not generating PSF maps will let the script run to completion.

Doing a simple dmcopy with the MASK filter with the strange binning (factor
of 3.216796875 derived from sky=min:max:#512), the failure can be seen from:

```
falafel-341: dmcopy test_maxsize_18701_broad_thresh.img"[sky=MASK(test_maxsize_18701_broad_thresh.expmap)]" \
/tmp/test_maxsize_mask.img verbose=5
The debug level is 5
Open
test_maxsize_18701_broad_thresh.img[sky=MASK(test_maxsize_18701_broad_thresh.expmap)] as table or image
CXC Data Model: DM V2.10
[*1]  Mask coordinate failure, mismatch 'LINEAR' offset for 'SKY'.

Failed to open virtual file
test_maxsize_18701_broad_thresh.img[sky=MASK(test_maxsize_18701_broad_thresh.expmap)]
# DMCOPY (CIAO 4.11): Mask coordinate failure, mismatch 'LINEAR' offset for
'SKY'.
```

2.) Another issue arises, seemingly from merging.py and header.py in ciao_contrib._tools that causes the script to fail when the 'psfmerge' parameter is set to 'exptime' or 'expmap' due to the merging rules.  ObsID 1595 only has ACIS-S3 active while the other observations have several chips active, so they don't have the same set of ONTIMEn/LIVTIMEn/EXPOSURn keywords, resulting in an attempt to sum keyword values of float and NoneType which throws an error.

In this particular case, removing the keywords in the event file works around the problem:

```
unix% dmcopy "acisf01595_repro_evt2.fits[ccd_id=7]" 1595_s3_repro_evt2.fits

unix% foreach f ( 2 3 5 6 8 )
? dmhedit 1595_s3_repro_evt2.fits filelist="" op=delete key=LIVTIME${f}
[...]

PR #268, opened 2019-09-11
missing exposure keywords set to 0 for CalcGTI merging rule

DougBurke:
This should allow the combination of PSF maps using exposure-related
scaling when one or more chips are missing in one (or more) obsids.

This should fix the second part of the problem seen in #262


DougBurke:
@kglotfelty - does it make sense that the calcGTI merging rule skips missing ccd values (so EXPOSUR3 of None, 300, 400 would end up with a merged value of 700)?

PR #269, opened 2019-09-11
Update instrument-map weight code to make more use of public interfaces

### LINKED ISSUE #258: update needed to make_instmap_weights change for sherpa i/f
I'm seeing a failure in tests

    # make_instmap_weights (06 June 2019): ERROR 'Data1DInt' object has no attribute '_get_mask'

I'm assuming this is related to the object refactoring in sherpa 4.11.1

DougBurke:
CIAO 4.12 removes the _get_mask method from the Data hierarchy, but prior
to that we can use the mask attribute to get at the values. This change
should not break CIAO 4.11 or 4.12 behavior, but there's a lot of needed
changes to the Sherpa data hierarchy in CIAO 4.12 so there may be other
problems.

This is to fix #258

### DIFF
--- a/ciao-4.11/contrib/lib/python3.5/site-packages/sherpa_contrib/utils.py
+++ b/ciao-4.11/contrib/lib/python3.5/site-packages/sherpa_contrib/utils.py
@@ -189,7 +189,7 @@ class InstMapWeights:
 
         # Apply the filter if necessary.
         #
-        m = data._get_mask()
+        m = data.mask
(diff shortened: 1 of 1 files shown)


DougBurke:
Rebased to pick up recent changes.

DougBurke:
Another "fix" that doesn't need a changelog. I'm pulling it in now (rather than for CIAO 4.12) as it should be a "no op" change.

PR #278, opened 2019-09-25
chandra_repro make sure set_verbosity set before using logger

### LINKED ISSUE #276: chandra_repro using logger before verbosity set
`chandra_repro` is using the `v1` logger before `set_verbosity` has been called.  In Python 3.7 this is causing all subsequent calls to `v1` to not print anything -- even after `set_verbosity`

We need to re-arrange the clode a little to make sure we call `set_verbosity` first.  Basically just move this

https://github.com/cxcsds/ciao-contrib/blob/c2d9c0721203c6a4add7f28d71c1b437fa898626/ciao-4.11/contrib/bin/chandra_repro#L2262

after

https://github.com/cxcsds/ciao-contrib/blob/c2d9c0721203c6a4add7f28d71c1b437fa898626/ciao-4.11/contrib/bin/chandra_repro#L446

kglotfelty:
This fixes #276 

Help file is updated, I don't think we need anything in change log (no diff to user)

### DIFF
--- a/ciao-4.11/contrib/bin/chandra_repro
+++ b/ciao-4.11/contrib/bin/chandra_repro
@@ -34,7 +34,7 @@ Aim:
 """
 
 toolname = "chandra_repro"
-version = "11 Jun 2019"
+version = "24 September 2019"
 
 # import standard python modules as required
 #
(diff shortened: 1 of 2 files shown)


DougBurke:
I do wonder if people might read the changes section and go "oh, what different version of python", but then again it's almost December!

Issue #190, opened 2018-11-30
possible bug, sort order of interleaved mode obi in merge-obs


kglotfelty:
https://github.com/cxcsds/ciao-contrib/blob/df6b0970dd7c221a643cfcacf0c4227433e081b5/ciao-4.11/contrib/lib/python3.5/site-packages/ciao_contrib/_tools/merging.py#L233

Sorry for the vague report ...

I sometimes see the `merge_obs` regression test report diffs in the interleaved mode tests.  It seems related to whether the files are sorted _e1_ then _e2_ or visa versa.  I'm not sure if the above snipet is the place where this sorting happens  or not.  That is both evt files will have the same MJD_OBS so I'm not sure if the sort order is unique.

Nominally this doesn't really matter -- except, the DTCOR in the e1 and e2 are very different, and so would affect the final LIVETIME and EXPOSURE time.  Note: this is somewhat related to #177 

FWIW: I'm seeing this on Ubuntu, and it, itself, does something funky with sort (the default sort order isn't the same as CentOS, I need to set LC_ALL=C to get defacto-standard sort order).  So I don't know if that also affects how python is sorting.

I'll post an example later -- I just wanted to capture now.

DougBurke:
I'm going to assume #286 fixed this, but please feel free to re-open it if the regression tests still fail !

PR #286, opened 2019-10-21
merge_obs: fix sort order when dealing with interleaved mode

kglotfelty:
`merge_obs` (and I guess `reproject_obs`?) sorts the input list of event files based on `TSTART` .

But interleaved mode datasets have the same `TSTART` so the e1 and e2 files can end up in a random order.  This affects things like DTCOR, LIVETIME, etc keywords, and which events show up first in the merged event file.

This change adds appends a sortable token (string) to the end of the TSTART to ensure a deterministic order.

I've chosen to sort so that the e2 file comes first since it is the longer of the two -- that is most events come from the e2 dataset so it's DTCOR/etc are appropriate for more of the events.

## TODO

- Review whether we should actually `dmsort` the event file on `TIME` if it contains both `e1` and `e2`.

### DIFF
--- a/ciao_contrib/_tools/merging.py
+++ b/ciao_contrib/_tools/merging.py
@@ -1076,9 +1076,8 @@ def validate_obsinfo(infiles, colcheck=True):
         # Sort primary second, secondary (longer) first, HRC is None.
         sort_order = { 'P' : 2, 'S' : 1, None: 0 } 
 
-        # leading zero pad to preserve numeric sort order.
         sort_tag = sort_order[obs.obsid.cycle]
-        time_tag= "{:020.3f}_{}".format(obs.tstart,sort_tag)
+        time_tag= (obs.tstart,sort_tag) # tuples sort too
         obsinfos.append((time_tag, obs))
 
     ninfiles = len(obsinfos)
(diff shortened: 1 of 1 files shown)


kglotfelty:
This will close #190 if/when completed.

DougBurke:
I haven't looked at the code yet (technically still on "vacation" this morning) but are there any more code changes, or is it WIP for your "do you even dmsort" question?

kglotfelty:
This should be "it" now.  The other sort fileio is the `MJD_|-OBS` thing.

The secondary -- should the file be `dmsort`ed is for another day

DougBurke:
Is it worth adding to the documentation that interleaved mode is now handled correctly?

kglotfelty:
I'm not sure "correctly" is correct.  It's now deterministic ; sometimes e1 would come first some times e2 would be first -- this just ensures e2 is first.

Why e2?  This is the same issue when combining any files where DTCOR is different -- just in the interleaved mode case instead of 0.99 vs. 0.98 kinds of diffs, it's 0.9x vs. 0.09 (ie an order of magnitude) ... what is the correct DTCOR in the merged dataset (and then LIVETIME/EXPOSURE).  e2 is the longer exposure so most events are e2 so the e2 DTCOR is probably "better" ... "correct" is a bit more of a leap.

DougBurke:
"Is it worth adding to the documentation that interleaved mode is now handled consistently?"

PR #287, opened 2019-10-22
csc search scripts almost final updates for csc2

kglotfelty:
This provides the param interface updates to search scripts

The final release token will need to be changed before we release `rel2.0` (TBD)


DougBurke:
Ta. I guess we really should try adding ahelp examples showing csc2 in action, but that can be done at a later date.

PR #289, opened 2019-10-24
updated csc release versions

kglotfelty:
Fin.

We can look at adding sensitivity for Dec.


DougBurke:
So with this we can update the CHANGELOG file to say "CSC 2 support is available and used by default"?

Issue #295, opened 2019-11-05
deflare matplotlib deprecation warning


kglotfelty:
In the python 3.7 edition of conda ciao 4.12 `deflare` is printing the following deprecation warning:

> INSTALL/bin/deflare:172: MatplotlibDeprecationWarning: 
> The frameon kwarg was deprecated in Matplotlib 3.1 and will be removed in 3.3. Use facecolor instead.
>   plt.savefig(oplot, format=format, frameon=True)
> INSTALL/bin/deflare:172: UserWarning: Tight layout not applied. tight_layout cannot make axes height small enough to accommodate all axes decorations
>   plt.savefig(oplot, format=format, frameon=True)
146a156,160

DougBurke:
Can this be closed now @nplee?

Issue #293, opened 2019-11-04
lc_clean plots with matplotlib 3.x.x


kglotfelty:
Just putting this here so it's not forgotten --

With matplotlib 3 , the lc_clean plots are coming out odd:  squashed with lots of white space between them:

![deflare2](https://user-images.githubusercontent.com/6662179/68126698-585c0380-fee2-11e9-81f7-19f932db3db5.png)

when I'd looked at this earlier I was seeing warnings like:

    /data/lenin2/Test/CIAO4.12/Conda/miniconda/envs/ciao-4.12b1/lib/python3.5/site-packages/matplotlib/tight_layout.py:209: UserWarning: tight_layout cannot make axes height small enough to accommodate all axes decorations
      warnings.warn('tight_layout cannot make axes height small enough '

This was only happening in the conda build but now in 4.12b2 where we've updated classic ciao it's showing up there too.

DougBurke:
Can this be closed now @nplee?

nplee:
yep

PR #298, opened 2019-11-05
updated lightcurves.py and deflare to handle matplotlib 3 deprecations

nplee:
these changes should address the deprecation warnings and drastically different plots reported in #293 and #295 with matplotlib 3.


ciao-4.11/contrib/lib/python3.5/site-packages/lightcurves.py
```diff
@@ -566,7 +566,17 @@ def plot(self, rateaxis="y", gcol="lime", erase=True):
         # setup figure
 
         if rateaxis == "y":
-            fig, axs = plt.subplots(2, 1, tight_layout=True)
+            from matplotlib import tight_layout
```
DougBurke: @nplee - this still works with matplotlib 2 ? I'm guessing so but just want to check.

nplee: yes it seems to work with matplotlib 2 from my limited testing.

PR #305, opened 2019-11-13
Handle numpy div by zero, missing ASCDS_CONTRIB env variable, MJD-OBS/MJD_OBS in fileio

### LINKED ISSUE #257: MJD?OBS keyword updates
Need to handle both `MJD_OBS` or `MJD-OBS` keywords

- [ ] `fileio` module

Grepping I also see it in `merge_obs` and `fluxobs` but it looks like just a comment.

### LINKED ISSUE #296: merge_obs numpy 1.15.x warning
There is a new numpy warning in the conda-ciao-py3.7 edition from merge_obs

> INSTALL/lib/python3.7/site-packages/ciao_contrib/_tools/merging.py:2592: RuntimeWarning: divide by zero encountered in true_divide
>   newvals = numerator / denominator

in the wavdetect_merged thread. I'm assuming we just need to add a `divide='ignore'` to `np.seterr()` ?

```python
  2590      res = np.seterr(invalid='ignore')
  2591      try:
  2592          newvals = numerator / denominator
  2593          newvals = newvals.astype(np.float32)
  2594      finally:
  2595          np.seterr(**res)
```

FWIW:

    ciaot: python -c 'import numpy;print(numpy.__version__)'
    1.12.1

    conda: python -c 'import numpy;print(numpy.__version__)'
    1.15.2

### LINKED ISSUE #304: assumption of paths with conda builds: ASCDS_CONTRIB
`ciao_contrib/_tools/run.py` contains the following, which is going to fail in conda builds when there's no `$ASCDS_CONTRIB` environment variable.

```
    v3("Looking for the lookup table for: " + toolname)
    for f in [dirpath + "/../data/" + fname,
              dirpath + "/" + fname,
              os.getenv("ASCDS_CONTRIB") + "/data/" + fname,
              os.getenv("ASCDS_CALIB") + "/dmmerge_header_lookup.txt"]:

        f = os.path.normpath(f)
        v3("Looking for: " + f)
        if os.path.exists(f):
            return f
```

DougBurke:
- fix #257 for `ciao_contrib._tools.fileio`

The following only address the merge/flux suite of scripts.

- fix #304 - `ASCDS_CONTRIB` is not guaranteed to be defined (make ASCDS_CALIB optional too for "safety")

- fix #296 - hide divide-by-zero warnings seen with recent NumPy versions.

### DIFF
--- a/ciao_contrib/_tools/fileio.py
+++ b/ciao_contrib/_tools/fileio.py
@@ -953,36 +953,52 @@ def expand_evtfiles_stack(instack, pattern="*evt*"):
 # Do we need to use MJD; can we not just use TSTART?
 #
 def sort_mjd(filenames):
-    "Return filenames, sorted by MJD_OBS, as a list"
+    """Return filenames, sorted by MJD_OBS or MJD-OBS, as a list
+
+    The keyword used to be MJD_OBS but was changed to MJD-OBS
+    circa CIAO 4.12.
+    """
 
     if len(filenames) < 2:
         return list(filenames)
 
     vals = []
-    dmkeypar = runtool.make_tool('dmkeypar')
+
+    # This used to use dmkeypar to get the keyword, but with the
+    # need to support multiple keywords it was changed to
+    # access the keywords directly in CIAO 4.12.
+    #
     for filename in filenames:
-        v3("Looking for MJD_OBS keyword in {}".format(filename))
+        v3("Looking for MJD-OBS keyword in {}".format(filename))
+
+        hdr = get_keys_from_file(filename)
         try:
-            dmkeypar.punlearn()
-            dmkeypar(infile=filename, keyword='MJD_OBS', echo=False)
-            mjdobs = dmkeypar.rval
+            mjdobs = hdr['MJD-OBS']
+        except KeyError:
+            v3("Not found, looking for MJD_OBS")
+            try:
+                mjdobs = hdr['MJD_OBS']
+            except KeyError:
+                v3("Also not found, so calculating MJD_OBS value.")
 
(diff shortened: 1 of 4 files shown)


ciao_contrib/_tools/fileio.py
```diff
@@ -953,36 +953,52 @@ def expand_evtfiles_stack(instack, pattern="*evt*"):
...
+        except KeyError:
+            v3("Not found, looking for MJD_OBS")
+            try:
+                mjdobs = hdr['MJD_OBS']
+            except KeyError:
+                v3("Also not found, so calculating MJD_OBS value.")
 
-        except IOError:
-            v3("MJD_OBS keyword missing, so calculating MJD_OBS value.".format(filename))
+                # MJD(TT) = MJDREF(TT) + ( TIMEZERO + TIME ) / 86400
```
kglotfelty: As far as the change itself goes, I think it looks OK.  

But -- I just wonder if it would simplify things if we just ignored `MJD_|-OBS` and just use this fallback calculations?  Something to think about; not a lien on this PR.

FWIW: I am still worried we will run out of precision using doubles for `TIME` values and will need to change `TIMEZERO` at some point in which case just sorting on  `TSTART` won't be sufficient.

DougBurke: I have a note for this routine that asks "can we not just sort on TSTART"? I think @nplee wrote this originally - can you remember why we sort on MJD-OBS here rather than some other time system?

DougBurke: I'm going to merge this now, since we can think about changing the routine in another PR.

PR #310, opened 2019-11-14
tgsplit bugfix for tgextract2 files

### LINKED ISSUE #294: tgsplit problem with tgextract2 files
I thought we'd gone through this but `tgsplit` needs to check for some columns in the source spectrum before it tries to delete them:

https://github.com/cxcsds/ciao-contrib/blob/96a606e64dac02143c1eb3e2c3e369132e0bb51e/ciao-4.11/contrib/bin/tgsplit#L222-L237

Since tgextract2 doesn't have `bg_err` column, this throws an exception.  We did this for the background but not the source.

kglotfelty:
This fixes #294  -- exception when trying to delete non-existent columns

### DIFF
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -19,6 +19,12 @@ Updated scripts
 
     The script has been updated to work with matplotlib 3.1.
 
+  tgsplit
+  
+    Fixes a bug when trying to work with TYPE:II pha files created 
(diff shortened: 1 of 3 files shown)


DougBurke:
Can you add an entry into the changes file for this?

kglotfelty:
done.

PR #314, opened 2019-11-15
Remove chips_contrib code and use of chips in ahelp documentation

DougBurke:
This removes the `chips_contrib` module hierarchy, related ahelp files, and ChIPS support from the Sherpa radial-profile code.

The `add_fov_region` and `add_ds9_contours` functions could be re-written to work with matplotlib, but for now they have been removed.

The documentation for the `sherpa_contrib.profiles` preference commands (e.g. `get_data_prof_prefs`) has commented out the tables listing the default settings as it's a lower priority to get this changed to mpl settings.


DougBurke:
@kglotfelty @nplee - just a heads up about this change. It shouldn't break SDS regression tests (since they would have already been broken by the removal of ChIPS).

PR #326, opened 2019-11-20
Support for VFAINT/no PHA cleaning and particle rate scaling method

nplee:
support for VFAINT observations without PHA filtering (acis_process_events run with check_vf_pha=no) added and new scaling method based on high-energy particle rate.

 Updated 'blanksky' ahelp for 'particle-rate' weight method and updated 'blanksky_sample' to handle 'particle-rate' scaling.

### DIFF
--- a/bin/blanksky
+++ b/bin/blanksky
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #
-#  Copyright (C) 2018, 2019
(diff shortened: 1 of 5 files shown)


DougBurke:
@nplee - so this is ready to go now?

nplee:
@DougBurke I just re-reviewed the changes here, and they're all set.  I just need to add new regression tests in our suite now.

PR #331, opened 2019-11-27
check_ciao_version: handle contrib installed manually

DougBurke:
This avoids the unhelpful failure message that check_ciao_version
returned if the contrib code has been installed via 'python
setup.py install' (which is only expected to be found for internal
testing).

The code had been written to handle the error case, but I had not
taken advantage of this before. So there's two fixes:

  a) correct handling of the error case [this was from an old version of the code; it has been subsumed by other changes]
  b) skip checking for any CIAO packages installed via pypi

The ahelp file has been updated to make conda the "primary" installation method (i.e.mentioned first) and ciao-install second.

I've also taken advantage of f-strings.

### DIFF
--- a/bin/check_ciao_version
+++ b/bin/check_ciao_version
@@ -50,13 +50,11 @@ Aim:
 
 import os
 import sys
-import glob
+
 from optparse import OptionParser
 
-import socket
-
 toolname = "check_ciao_version"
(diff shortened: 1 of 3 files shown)


DougBurke:
@kglotfelty - this should fix the problem you reported to me in our regression tests.

PR #338, opened 2020-01-21
dax: psffrac hanging problem

### LINKED ISSUE #255: dax get_psffrac script when user pref is arcmin
https://github.com/cxcsds/ciao-contrib/blob/1383aa47f55e835d8b05f79804810e0575128561/ciao-4.11/contrib/bin/ds9_get_psffrac.sh#L55

When users like @nplee have their preferences set to display regions in arcmin (instead of degrees), this breaks the get_psffrac script because the string get passed in with a trailing tick `'` which causes and illegal python syntax which results in the script hanging.

I couldn't find a way to control the region pref (degrees|arcmin|arcsec) when making the above xpa call (the `degrees` there are opposed to `sexagesimal`.

I'll need to check for the tick and scale to degrees in the script itself.

kglotfelty:
This fixes #255 

Originally I thought it was a problem with preferences on units but now realize it was a diff in pref in format (ciao vs. ds9 region format).  Dax should ask for format it wants instead of relying on defaults.

### DIFF
--- a/share/doc/xml/dax.xml
+++ b/share/doc/xml/dax.xml
@@ -90,9 +90,10 @@ unix% ds9 -analysis $ASCDS_INSTALL/contrib/config/ciao.ds9 ...
 <ADESC title="Changes in the 4.12.2 (March 2020) release">
 
   <PARA>
-    Updated the Regions -> PSF Fraction task to be sure to request 
-    regions in ds9 format. If user have changed default to 'ciao' 
-    format, this task will hang.
+    Updated the Regions -> PSF Fraction task to force regions in
+    ds9 format.  This update prevents dax from hanging when 
+    users have modified their preferences to save regions in 
+    ciao format.
   </PARA>
 
 </ADESC>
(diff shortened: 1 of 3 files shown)


share/doc/xml/dax.xml
```diff
@@ -87,6 +87,19 @@ unix% ds9 -analysis $ASCDS_INSTALL/contrib/config/ciao.ds9 ...
 </DESC>
 
 
+<ADESC title="Changes in the 4.12.2 (March 2020) release">
+
+  <PARA>
```
DougBurke: I think this message (and the one in the changelog) reads like there is still the chance of there being a hang. Perhaps change the second sentence to

```
This stops the task from hanging if users have changed the default region format to 'ciao' format.
```

Or I mis-understand things again.

kglotfelty: Is the latest better?

[minor fight w/ git trying to resolve merge in Changes file ]

Issue #330, opened 2019-11-26
ftp goes bye bye


DougBurke:
What is going to be affected once the CXC FTP sites are turned off?

 - `download_chandra_obsid`
 - library functions that are called by `download_chandra_obsid`

Anything else?

kglotfelty:
`download_obsid_caldb` too

DougBurke:
This has been fixed in #340 and #347

PR #340, opened 2020-01-29
download_obsid_caldb updates

kglotfelty:
This has two updates

- Now using https:// rather than ftp:// , part of #330 
- Now creates standard caldbinit.unix and caldbinit.sh setup scripts in outdir.

### DIFF
--- a/bin/download_obsid_caldb
+++ b/bin/download_obsid_caldb
@@ -24,12 +24,13 @@ import os
 import sys
 
 from urllib.parse import urlparse
+import urllib.request 
 
 import ciao_contrib.logger_wrapper as lw
 
 
 toolname = "download_obsid_caldb"
-__revision__ = "18 June 2019"
+__revision__ = "29 January 2020"
 
 lw.initialize_logger(toolname)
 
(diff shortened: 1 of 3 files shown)


DougBurke:
I was going to ask how much of this could be pushed up into library routines that I could just use in `download_chandra_obsid`, but it looks like not much.

kglotfelty:
Not really much.  The getting the `Content-Length` from the header and reading in chunks so we can get the progress bar would be about it; but that's a pretty small bit-of-code.

In the ftp version I used the `LIST` command to get the list of subdirs (and the filtered out the know list of things to avoid).  In this version, I just put in the list of dirs I know I need so I didn't need `LIST` anymore (so no need to parse dir. listings).

bin/download_obsid_caldb
```diff
@@ -226,27 +79,30 @@ def retrieve_config_files(ftp, newver, rootdir, outdir):
...
-        out_indx = open("{}/config/{}-{}.indx".format(outdir, dd, newver), "wb")
-        out_key = open("{}/config/{}-{}.config".format(outdir, dd, newver), "wb")
+        out_indx2.close()
         verb2("Index name :" + out_indx.name)
-        verb2("Key Config name :" + out_key.name)
-        ftp.retrbinary("RETR caldb.indx", out_indx.write)
-        ftp.retrbinary("RETR key.config", out_key.write)
-        out_indx.close()
+
+        out_key = open("{}/data/chandra/{}/key.config".format(outdir, dd), "wb")
```
DougBurke: Would it be best to do the download first - or at least the connection - and then open the files (in case of an error we don't get an empty file left around, assuming that the open command on its own actually creates an entity and isn't just lazy)?

So something like (where `name1` and `name2` are the unversioned/versioned name?

```
p = ...urlopen().read()
for n in [name1, name2]:
  with open(...) as out:
    out.write(p)
```

DougBurke:
We should probably set the user agent string (even if it is going away at some point) to make it easier to track downloads via our scripts.

bin/download_obsid_caldb
kglotfelty: latest version `read`s these files before writing.

kglotfelty:
added User-Agent w/ simple wrapper.

PR #342, opened 2020-02-03
Object-oriented wrapper for CIAO PSF module

hamogu:
The CIAO psf module gives access to the CALDB REEF file from Python.
However, it's use is a little cumbersome: It requires several lines of
boilerplate code and two imports, just to return one number.
Additionally, the user needs to remember to free resources.

This module warps psf in a OO interface available as a context manager.

I'm opening this PR with two questions:

- Is this even in the scope of ciao-contrib?
- Is this the right directory to put stuff like this?

If the answer to the questions above is "yes", then I still have to:

- [ ] write docstrings

### DIFF
--- a/ciao_contrib/psf_contrib.py
+++ b/ciao_contrib/psf_contrib.py
@@ -83,8 +83,8 @@ def psfFrac(energy, theta, phi, size):
     eef : float
         enclosed count fraction
     """
-    with PSF() as ecf:
-        return ecf.psfFrac(energy, theta, phi, size)
+    with PSF() as _psf:
+        return _psf.psfFrac(energy, theta, phi, size)
 
 
 def psfSize(energy_keV, theta_arcmin, phi_deg, ecf):
@@ -108,5 +108,5 @@ def psfSize(energy_keV, theta_arcmin, phi_deg, ecf):
     eef : float
         enclosed count fraction
     """
-    with PSF() as ecf:
-        return ecf.psfSize(energy_keV, theta_arcmin, phi_deg, ecf)
+    with PSF() as _psf:
+        return _psf.psfSize(energy_keV, theta_arcmin, phi_deg, ecf)
--- a/ciao_contrib/psf_contrib.py
+++ b/ciao_contrib/psf_contrib.py
@@ -1,0 +1,112 @@
+#
+#  Copyright (C) 2020
+#            Smithsonian Astrophysical Observatory, MIT
+#
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 2 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


kglotfelty:
So -- not to answer your questions, but just tried it and ran into a problem importing in python3.5


```python
% ipython
Python 3.5.6 |Anaconda, Inc.| (default, Aug 26 2018, 21:41:56) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.5.0 -- An enhanced Interactive Python. Type '?' for help.
f
In [1]: from ciao_contrib.psf_contrib import *
AttributeError                            Traceback (most recent call last)
<ipython-input-1-f1fb68e0ef95> in <module>()
----> 1 from ciao_contrib.psf_contrib import *

/data/lenin2/Test/CIAO4.12/Conda/miniconda/envs/hack35/lib/python3.5/site-packages/ciao_contrib/psf_contrib.py in <module>()
     40 
     41 
---> 42 class PSF(contextlib.AbstractContextManager):
     43     def __init__(self, pdata=None):
     44         if pdata is None:

AttributeError: module 'contextlib' has no attribute 'AbstractContextManager'
```

hamogu:
I develop on 3.7 and I promise it works there. If we want something like this at all, then I'll make sure
- [ ] it works on all supported versions
- [ ] test the numerical results for more paramters
- [ ] add docs

However, if you think this is out of scope for ciao-contrib then I won't bother with those points and just continue to use it in my own code, which uses 3.7 and where I tested it for the limited range of parameters that I need there.

DougBurke:
@kglotfelty - is the PSF library one of those "singleton"/"Highlander" style libs we have (there can only be one instance and if you try to have multiple times you'll end up with a series of terrible follow-up movies, err I mean potential memory oopses)? Or is it safe to wrap up like this?

kglotfelty:
I do not think so -- I think this is safe.   We (ie me wearing my old hat) learned what not to do based on `pixlib` ; the psflib I believe is free of globals that could cause grief.

hamogu:
So is the answer that you would accept this for CIAO 4.13 (when we drop 3.5 anyway) if I add some docs before?

DougBurke:
@hamogu - sorry I've just been shirking on this. What do you think needs to be done here? We are only interested in python 3.6 - 3.8 now. Unfortunately we don't have much time to finalize if we want this in the 4.13 release (roughly one more week).

It looks like it has enough docstrings in it to be usable (even if more wouldn't harm)? We could add ahelp files, but it's not required.

hamogu:
I had already forgotten about it, and just now read the code again, so that should almost count as an independent review ;-)
I suggest to merge it. I think it's good to add to the contrib and in fact I'm using this code in my contributions to Dave P.s HETG confusion code with eventually will make it into the contributed scripts as well. Having this in 4.13 would make that part easier. While I could write more docs, I'm not sure now what I had in mind. This is not going to be widely used, so it does not need an ahelp file, I think.
The fact that I used it there for a while with no problems coming up also give me confidence that no test are needed. After all, this just wraps CIAO and does not do an calculations by itself.

So, please merge or let me know what I should do.

DougBurke:
We'll need something added to the changelog but I can add that

PR #343, opened 2020-02-03
Fix example in XML helpfile for coords_chandra

hamogu:
As given, the dictionary created in the example would not be `{header_key: header_value}`, but `{crates_file_object: header_value}` which is clearly a typo in the example.


DougBurke:
Ta

PR #348, opened 2020-02-06
download_obsid_caldb set unverified ssl context

kglotfelty:
Thanks to the reminder from @DougBurke , we need to set the ssl context so that the ciao-install version of python works (certs are not configured).  

Tested on Mojave where it fails w/o this fix.


DougBurke:
I had considered doing this only as a fallback case, but then decided that it was tricky to get right.

Issue #341, opened 2020-02-03
correct_periscope_drift doesn't support verbose=0


kglotfelty:
```bash
% correct_periscope_drift mode=h --debug verb=0
# correct_periscope_drift (0.3): ERROR 0


##### Traceback:

Traceback (most recent call last):
  File "/export/ciao-4.12/contrib/lib/python3.5/site-packages/ciao_contrib/logger_wrapper.py", line 531, in new_fn
    return fn(*args, **kw)
  File "/export/ciao-4.12/contrib/bin/correct_periscope_drift", line 476, in main
    logger.setLevel(LOGLEVELS[opt['verbose']])
KeyError: 0
```

PR #349, opened 2020-02-06
correct_periscope_drift changes

DougBurke:
- [X] support verbose=0 (#341)
- [X] hide python traceback parameter validation fails
- [X] remove ChIPS code
- [X] use lower-level Sherpa routines as do not need the session management

### DIFF
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -42,6 +42,9 @@ Updated scripts
     Fixed bug that caused parameter values in the script's HISTORY 
     records to be reset to their default (blank) values.
 
+  correct_periscope_drift
+
+    The script will now run when the verbose parameter is set to 0.
 
 ## 4.12.1 - December 2019 ##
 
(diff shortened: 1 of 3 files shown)


DougBurke:
@kglotfelty @jeanconn - this started as a clean up to support verbose=0 but then I decided to go and change-up the Sherpa code so that it uses the "low-level" interface. This turns out to be another way to fix the verbose=0 issue (as the UI layer creates the screen output we are trying to hide, we no longer need to hide it now we're not using the UI layer. Yay).

I haven't taken the next step and cleaned things up a bit more (I'm sure there's something that could be done).

If you could look this over (and test a bit) that would be appreciated. I realize neither of you likely know the low-level Sherpa API, but any comments appreciated. I tested one random obsid and didn't see a difference, but I don't really know where "problem" cases are.

jeanconn:
Hi @DougBurke .  Sorry late to the party.  This looks fine to me and makes sense.  With regard to testing, I don't bring much extra... our previous example from the docs still looks reasonable.  Should probably add a grid to the plots at some point.

PR #350, opened 2020-02-07
More helpful error for choosing wrong pix_adj parameter

hamogu:
A local user just asked me for help after choosing `pix_adj="edser"` and receiving an "invalid pix_adj" error. I then started to look into the code to see where it came from and only after I realized that the ObsID in question was HRC data. Since the error messages are thrown inside an if block that tests for the instrument name, it can't hurt to print out the instrument name in the error message. When working with just one observation, users should know what they are dealing with, but anyone who processes a number of ObsIDs might not automatically associate ObsID and instrument for all datasets and might appreciate the hint.

### DIFF
--- a/bin/chandra_repro
+++ b/bin/chandra_repro
@@ -560,7 +560,7 @@ def event1_inputs(params):
                 elif pix_adj_value == 'edser':
                     params["pix_adj_value"] = 'EDSER'
                 elif pix_adj_value != 'default':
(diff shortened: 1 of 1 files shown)


DougBurke:
@kglotfelty - these LGTM but can you sign off on them before we merge them?

DougBurke:
@hamogu  - Sorry for the delay in merging.

PR #352, opened 2020-02-12
srcflux updates to use new skyfov algo.

kglotfelty:
Tested w/ standard regress test -- all used `ASPSOL` as expected.

Ran data provided by Joe for 10.8.3 testing; confirm that `skyfov` ran w/ `method=convexhull` and no errors.  (No diffs either for data in hand.)

### DIFF
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -46,6 +46,13 @@ Updated scripts
 
     The script will now run when the verbose parameter is set to 0.
 
+  srcflux
+  
+    Uses the new skyfov method=convexhull algorithm when new
+    aspect solution files with CONTENT=ASPSOLOBI are used.
+
+
+
 ## 4.12.1 - December 2019 ##
 
 Updated scripts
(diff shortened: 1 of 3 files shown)


DougBurke:
Should we have a parameter that lets a user select the old `method` value for `skyfov` (in this and other tools)? If they are in the middle of an analysis (maybe multiple datasets) then they may not want to change thing.

This is more an issue for other scripts (e.g. `fluximage` / `merge_obs`) but I thought we could/should make a policy decision.

kglotfelty:
I don't think so.

The fov file they get from the archive will use method=convexhull when the new ASPSOLOBI files are produced.

We never want to let users use method=convexhull if using old (current) ASPSOL files -- the box is too tight for observations with large DY,DZ offsets.

There isn't really any advantage to using method=minmax with the new ASPSOLOBI files.  The sky cooridnates are all going to be different anyway (eg for the repro-5 datasets) so it's not like you'll be able to compare them easily.

DougBurke:
Now that DS 10.8.3 is out, is this good to go or does it still need testing against data "in the wild"?

PR #353, opened 2020-02-12
chandra_repro updates to use skyfov.method=convexhull for new asol files

kglotfelty:
Not sure the DS10.8.3 blurb belongs in the ahelp file but needs a little bit of context methinks.

### DIFF
--- a/bin/chandra_repro
+++ b/bin/chandra_repro
@@ -1,13 +1,13 @@
 #!/usr/bin/env python
 
 #
-# Copyright (C) 2010-2019  Smithsonian Astrophysical Observatory
+# Copyright (C) 2010-2020  Smithsonian Astrophysical Observatory
 #
 #
 #
 # This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


bin/chandra_repro
```diff
@@ -2137,6 +2137,34 @@ def set_ardlib(params):
     v2("Observation-specific bad pixel file is set in ardlib.par.")
 
 
+def choose_skyfov_method( asolfiles ):
```
DougBurke: We (probably me) should look at the flux/repro image/obs stuff, since I think there's code there that runs skyfov. Perhaps it should be consolidated?

DougBurke: It looks like very-similar (if not the same) functionality is used in `srcflux` too (looking at #352 )

kglotfelty: They are similar but a little different.   Small enough snipet of code that adding the conditionals to pick a|b didn't seem worth it to me.

Issue #344, opened 2020-02-03
dax funcounts bugs


kglotfelty:
If there are multiple ds9's running, the dax Net Counts (aka funcounts) script behaves badly -- it need to check for multiple instances and exit nicely w/ error message.

It also needs to be sure it's getting physical coords.  This should be the case already, but Antonella and her student came across a situation where it was getting passed radii and arcmin|arcsec; the `'` or `"` didn't work-and-play well with the runtool|param stuff.

Also, RFE: include the `ERR_RATE` column from dmextract to the output

Issue #311, opened 2019-11-14
dax fit plots match sherpa


kglotfelty:
@DougBurke , based on sherpa plotting changes like https://github.com/sherpa/sherpa/pull/667 I'm guessing there's an implicit RFE to have the dax plots try to plot model over data ?

DougBurke:
If you try with the chips backend then I think you'll see that 'plot_fit' has model over data - and 667 was an attempt to fix this with matplotlib (due to changes in mpl v2) - e.g. https://github.com/sherpa/sherpa/pull/667#issuecomment-540062240

Which is a long-handed way of saying "probably yes", but I haven't seen what the current dax plots look like, so can't say for sure.

Issue #299, opened 2019-11-06
dax move plotting code into imported module


kglotfelty:
The dax spectral fit script writes what was a short python script to run sherpa, the `cmd` file.

With the ds9 plotting update all the plot and xpa commands are now including making the simple `cmd` file much larger and irrelevant since users will be plotting via `matplotlib`.

We should move all these plotting routines

https://github.com/cxcsds/ciao-contrib/blob/96a606e64dac02143c1eb3e2c3e369132e0bb51e/ciao-4.11/contrib/bin/ds9_specfit.sh#L270

into contrib/lib module so all users see in the `cmd` file is the `import` and a few `plot_foo` calls.

Issue #279, opened 2019-09-26
RFE: dax spectral fit expand model selection


kglotfelty:
Per AF:

We should expand the list of spectral models available via dax to match what's in [WebSpec](https://heasarc.gsfc.nasa.gov/webspec/webspec.html).

Also we should allow for a 2nd additive model component.

PR #356, opened 2020-03-05
dax updates for 4.12

kglotfelty:
This update includes fixes for #344 , #311 , #299 , #279 

- New: wavdetect can now be run via dax.  Users can supply exposure map and psfmap.
- New: grating coordinates/energy task 
- New: run simulate_psf (marx) for a monochromatic energy & flux.
- New: dmregrid2 to shift, rotate, scale image
- Updated: spectral fit now has all webspec models, 2nd additive model component.
- Updated: celldetect users can supply fixed cellsize or psfmap and expmap.
- Updated: simplified fit plots (model now plotted over data)
- Fix: problem in net-counts task (0 background from net-counts task), physical coords
- Fix: problem Nick found (hang when using psf-fraction task), also physical coords


DougBurke:
Is this ready for merging or do you want some testing first?

bin/ds9_imgproc_wrapper
```diff
@@ -178,9 +178,14 @@ elif "wavdetect" == toolname:
...
     tool.defnbkgfile=tool.outfile+"_nbkg"
     tool.interdir=os.environ["DAX_OUTDIR"]
+    tool.expfile=sys.argv[4][1:].strip('"') # strip off leading 'e' 
+    tool.psffile=sys.argv[5][1:].strip('"') # strip off leading 'p'
 elif "celldetect" == toolname:
     output_fn = send_output_region_to_ds9
     tool.fixedcell=int(sys.argv[3])    
+    tool.expstk=sys.argv[4][1:].strip('"') # strip off leading 'e' 
+    tool.psffile=sys.argv[5][1:].strip('"') # strip off leading 'p'
+    tool.thresh=float(sys.argv[6])
```
DougBurke: So, making the user pre-create these images is the simplest thing to do, but I wonder if it would make sense (for the dax-like workflow) to allow the task to create the exposure and/or psffiles on the fly? It would obviously complicate the interface (as users would want one of three options: 1) no file, 2) use a file, 3) auto-generate the file) and how to specify the needed parameters for the tools. As well as the life of whoever has to implement this.

Perhaps an enhancement for a future release?

kglotfelty: The problem is speed; exposure maps take too long at descent resolution to be done via dax.

I've asked Bill to give ds9 a 'File Open' dialog in the parameter selection gui (akin to what peg does).  That will help.

kglotfelty:
We can hold for now unless merging helps clean up any conflicts.  Not sure A will be able to do any testing before prop deadline.

bin/ds9_imgproc_wrapper
kglotfelty: It may also be useful for `fluximage` to put the filenames of the expmap and psfmap (if any) into header keywords of the counts image and fluxed image.

DougBurke: So then we'd also want to "invent" a syntax for these parameter values to say "use this keyword from the input file name" in a similar manner to the parameter libraries redirect/whatever functionality? Or would you just automatically use them if found (and the user didn't specify these arguments)?

kglotfelty:
probably just something like "INDEF" which already implies that the tool
should "do whatever it thinks is best"  vs. say "none" which explicitly
means, well "none"

DougBurke:
I'd assumed that INDEF was a numeric-only option (although technically I
guess as it's a string field you can make the special case be any string
you want, such as INDEF or "" or hey-baloney).

bin/ds9_imgproc_wrapper
DougBurke: Can you open an issue about coming up with some sensible keywords we can use for this?

kglotfelty:
I’m going to close this PR, more good stuff comin.

Issue #346, opened 2020-02-05
update skyfov method with new asol file?


kglotfelty:
When the new obi-asol files are released, we can|could|should think about updating the contrib scripts to use the new `method=convexhull` algorithm in `skyfov`.

A quick grep looks like

- [x] `chandra_repro`
- [x] `srcflux`
- [ ] `ciao_contrb._tools.fileio.py`

We can key off the `CONTENT` keyword

```bash
% dmkeypar 13819/primary/pcadf13819_000N001_asol1.fits content echo+
ASPSOLOBI

% dmkeypar 13819/secondary/pcadf444003563N001_asol1.fits content echo+
ASPSOL
```
Eg something like:

```python
choices={ 'ASPSOLOBI' : 'convexhull' , 'ASPSOL' : 'minmax' }
```

The advantage is that the FOV will, in general, be tighter (never larger).    Since we can key off values in data products we can make this change independent of new asol files in the wild.

Since there are more|variable number of points in the polygons not sure if that affects anything anywhere else?

DougBurke:
I think I use the skyfov to determine sky ranges for `fluximage`/`merge_obs`/..., so there may be some changes to the output data, but I would hope it doesn't materially change the data.

DougBurke:
With #358 having been merged I think we can close this now.

kglotfelty:
Correct.

PR #358, opened 2020-03-27
Support CONTENT=ASPSOL or ASPSOLOBI for skyfov runs: apply_fov_limits/fluximage,flux_obs,merge_obs

DougBurke:
Update code for DS 10.8.3 - new aspect solution lets us use skyfov method=convexhull (#346)

### DIFF
--- a/ciao_contrib/_tools/fileio.py
+++ b/ciao_contrib/_tools/fileio.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2019
+#  Copyright (C) 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2019, 2020
 #            Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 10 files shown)


DougBurke:
@kglotfelty - I've only been able to test this against HRC datasets with ASPSOLOBI / DS 10.8.3 data (both for `fluximage` and `reprojcet_obs/flux_obs` thanks to the Ar-Lac calibration observations. I'm not sure this makes a great test, but it seems to be okay.

ciao_contrib/_tools/fileio.py
```diff
@@ -306,22 +306,39 @@ def validate_asol(infile):
...
     # not just try a simplistic check.
     #
     (keys, cols) = get_keys_cols_from_file(infile)
+    try:
+        content = keys['CONTENT']
+    except KeyError:
+        raise IOError("The file {} is missing the CONTENT keyword.".format(infile))
+
+    if content not in ['ASPSOL', 'ASPSOLOBI']:
+        raise IOError("Aspect solutions have CONTENT=ASPSOL or ASPSOLOBI, but found {} in\n{}".format(content, infile))
```
kglotfelty, line 325: It's more a personal pref, but I'm not a big fan of multi-line error messages (the `\n` before the infile).    Just a pref.

PR #360, opened 2020-03-30
Update xspec verison and CIAO version in docs

hamogu:
Really we probably need a variable to update this in automatically on release. Or is it intentional that 4.11 is cited because that's when the tool was added?

TBD: Check if there are other places with the same text, but I'm not doing that until you tell me that we actually *want* to update.

Note: Other models did updates the XSPEC string and CIAO 4.12, e.g. https://cxc.harvard.edu/sherpa/ahelp/xspcfabs.html

### DIFF
--- a/share/doc/xml/load_xspartcov.xml
+++ b/share/doc/xml/load_xspartcov.xml
@@ -57,8 +57,8 @@
 	This information is taken from the
 	<HREF link="https://heasarc.gsfc.nasa.gov/docs/xanadu/xspec/manual/manual.html">XSpec
 	User's Guide</HREF>.
-	Version 12.10.0e of the XSpec
-	models is supplied with CIAO 4.11.
+	Version 12.10.1n of the XSpec
+	models is supplied with CIAO 4.12.
       </PARA>
 
     </DESC>


DougBurke:
So, I've mixed thoughts on this. The text hasn't been updated since it was added, so I haven't checked if the XSPEC documentation has been updated. So changing the version number is potentially problematic.

I'm also hoping that this particular part of the contrib code will not be needed with CIAO 4.13 (ie hoping https://github.com/sherpa/sherpa/issues/68 will be implemented). I've not concentrated much on this part of the contributed package recently because without the regrid code that is now in Sherpa many of these "convolution" models aren't that useful.

Technically the ahelp pages for XSPEC models provided as part of CIAO - such as your example of https://cxc.harvard.edu/sherpa/ahelp/xspcfabs.html - only say what version of XSPEC is being used, and not that the information provided in the page is from that version, but a reader isn't going to be happy with that technicality. The subtext of this comment is that the XSPEC version number is updated for the documentation provided as part of CIAO, even though there is no guarantee that the information on the page has been updated to match changes in the XSPEC documentation. So this would suggest it's okay to update the version info in the contributed documentation. 

Basically, this is currently a low priority for me but I'm willing to be convinced otherwise. With the recent-ish changes to the contrib code setup, it would be easier than in the past to add some form of document generating capabilities to automate this, but I'm not convinced it would be worth automating this given the scope of the problem.

hamogu:
How about we post-pone a decision until 4.13, when we know if the sherpa/sherpa#68 makes it or not? If we decide to go with a one time search and replace (I agree that that's probably good enough), then this is a very quick issue we can squeeze in in the beginning of December.

DougBurke:
Sounds good to me. Out of interest, did you notice this because you were using the model, or just happened to notice it?

PR #361, opened 2020-03-31
fix: find_chandra_obsid can now be given an obsid

DougBurke:
DS 10.8.3 added a blank line to the obsid output, and the script
wasn't smart enough to filter this out before checking the format.


DougBurke:
@kglotfelty - this should fix the problem. Can you just look at the change to make sure I'm not missing anything (there's the question of whether I should be doing as much of a check as I do, but that's for a different PR...)? Note that I snuck in a few unrelated cleanups; the bug fix is on line 508 - the creation of the `lines` variable.

Issue #357, opened 2020-03-09
possible RFE: specextract apply optional threshold to sky2tdet wmap


kglotfelty:
[I'm not sure this ever got circulated -- apologies if already considered.] 

Doing some spring cleaning in my home dir and I found this little snipet of code that truncates the sky2tdet wmap -- thus making mkwarf run faster.   We might want to think about adding  this as an option to specextract.  

I tried to attached but git doesn't like .py files as attachments. This is circa 2011 so Python2. 

```python
#! /usr/bin/env python
"""
The best way to deal with the increased run-time seen as a result of the
new sky2tdet / mkwarf combo for large regions is to threshold the wmap.

Rebinning the wmap is less desirable since it essentially can randomly
cause badpixels/columns/etc to be over or under weighted.  (It will be
the same every time, just random in that you don't have control over
it.)  If your region is huge (ie whole chip) then that probably won't
matter, if it's a modest size, eg off axis point like src, then binning
is bad.

One way to threshold (and there are several) is just to look at the
pixel values in the wmap, and determine a cutoff that preserves a
certain amount of the total flux. That way, if a badpixel/column/etc is
not weighted by a lot of flux it can be explicitly purged and likewise
if they are covered by a large flux, then they will be preserved and the
ARF appropriately weighted.

Attached is a script-let that does this.  It, or something like it,
might be worth considering adding to specextract.  (with appropriate
paramio, error/exception handling, and ciaoruntool-ified,etc)

Just doing some simple testing, for an off-axis point like source
covering ~4000 pixel, with a 5% threshold (so keep 95% flux), we can
eliminate over 40% of the pixels.  Replicating the recent helpdesk
user's problem of a background field (so single chip image w/ srcs cut
out), a 5% threshold takes out 12% of the pixels, 10% threshold -> 19%
pixels.  The more uniform the pixel values, the less percent area will
be suppressed.
"""

import os
import sys
import pycrates as pyc
import numpy as np

if ( len(sys.argv) != 4 ):
    print ("Usage: {0} infile outfile threshold\n".format(sys.argv[0]));
    exit(1)



infile=sys.argv[1]
outfile=sys.argv[2]
threshold=float(sys.argv[3])

#
# Read WMAP image and get pixel values
#
image=pyc.read_file(infile)
ii=image.copy_image_values()

#
# Sort pixel value
#
xx=ii.flatten()
xx.sort()

ii,=np.where(xx>0)
xx=xx[ii]



#
[...]

PR #365, opened 2020-04-01
specextract enhancements

nplee:
This update includes the use of the most recent HRC RMFs found in the CalDB and the weighted response speed up approach by clipping the sky2tdet WMAP as suggested in #357


DougBurke:
@nplee - I see this as two featires:

a) HRC RMFs

which are an "obvious" fix (although should we ask the CALDB what the file should be rather than hard-code the value)?

b) the #357 change

which is in the "need to evaluate" category (e.g. what fraction do we suggest to users, what differences does it make in the results, ...).

I haven't looked at the changes yet, but is the WMAP change on or off by default, and can you turn it "off" (ie get the same results as before the change) with a parameter?

nplee:
@DougBurke 

On (a), you're certainly correct that calquiz could be used to query the CALDB instead of the hardcoded RMFs; certainly a better option for future-proofing.

(b) The default behavior is not to apply the clipping, so same results as before. This is controlled by the "wmap_clip" parameter.

DougBurke:
@nplee - I'd like this separated into two: the RMF change and then the WMAP-filtering change. I only want to take the RMF change for this release, and we can have some time to check out the WMAP change. So, priority wise, just focus on a PR with the RMF change for now. Does this sound okay?

nplee:
@DougBurke yes, I'm working out a version with just the RMF changes.

nplee:
@DougBurke the RMF-only update can be found in PR #371

PR #367, opened 2020-04-02
WIP: update chandra repro for mix asol types

kglotfelty:
@DougBurke can you review this.    I still need to do ahelp|doc updates but wanted to get your thoughts.

Since we know we want users to use the aspsolobi files -- if they are there, then this just skips|omits the old files with a warning

```bash
% chandra_repro 22916 out=blek 

Running chandra_repro
version: 12 February 2020


Processing input directory '/stage/pool1/kjg/22916'

Warning: Original level 1 event file was created with asol1 file named '['pcadf22916_000N001_asol1.fits']', however, the file found on disk is named 'pcadf701242066N001_asol1.fits'.
WARNING: The primary directory has both CONTENT=ASPSOL and CONTENT=ASPSOLOBI files. Only the ASPSOLOBI file ['/stage/pool1/kjg/22916/primary/pcadf22916_000N001_asol1.fits\n'] will be used
Resetting afterglow status bits in evt1.fits file...

Running the destreak tool on the evt1.fits file...
...
```

There are actually 2 warnings ... the 1st because the ASOLFILE doesn't match the ASPSOL filename, and then the more explicit ASPSOL v. ASPSOLOBI warning.

Or, I can make that warning fatal.

### DIFF
--- a/bin/chandra_repro
+++ b/bin/chandra_repro
@@ -34,7 +34,7 @@ Aim:
 """
 
 toolname = "chandra_repro"
-version = "12 February 2020"
+version = "02 April 2020"
 
 # import standard python modules as required
 #
@@ -137,6 +137,63 @@ def error_out(params,msg,val):
     print("\n%s" % msg)
     raise ValueError("%s" % val)
 
+
+def filter_aspsolobi_from_aspsol(asol_list,mydir):
+    """As of DS10.8.3, there are now two flavors of aspect solution 
+    files, per OBI (CONTENT=ASPSOLOBI) and the original per-constant
+    aspect interval (cai) (CONTENT=ASPCOL).  Users should only ever
+    have one or the other. 
+    
+    Unfortunately, due to a quirk in the archive, users may have both.
+    
+    We know that there can be 1 and only 1 ASPSOLOBI file per obi (multi-obi
+    obsid's will need to have already been split, #repro5). 
+    
+    So if we find both ASPSOLOBI and ASPSOL files, we are going to pick
+    the ASPSOLOBI files.  This provides a better off-axis angle value.
+    
+    """
+    from pycrates import read_file
+
+    content_values = {}
+
+    for f in asol_list:
+        tab = read_file(f.strip("\n").strip(" "), mode="r")
+        content = tab.get_key_value("CONTENT")
+        if content in content_values:
+            content_values[content].append(f)
+        else:
+            content_values[content] = [f]
+        
+    if "ASPSOL" in content_values and "ASPSOLOBI" in content_values:
+        v0("WARNING: The primary directory has both CONTENT=ASPSOL and "+
+          "CONTENT=ASPSOLOBI files. Only the ASPSOLOBI file "+
+          "{} will be used".format(content_values["ASPSOLOBI"]))
+        # don't return yet, fall through to next check
+
+        # But I do want to go ahead and remove the old files from the
+        # output directory.
+        for oldasp in content_values["ASPSOL"]:
+            old = os.path.join(mydir, os.path.basename(oldasp.strip("\n").strip(" ")))
+            if os.path.exists(old):
+                os.unlink(old)
+
+
+    if "ASPSOLOBI" in content_values:
+        if len( content_values["ASPSOLOBI"] ) > 1:
+            error_out("Too many ASPSOLOBI aspect file found.  If this "+
+            "is a mutli-obi observations, be sure to split it first "+
+            "using the splitobs script")
+        return content_values["ASPSOLOBI"]
+    
+    if "ASPSOL" in content_values:
+        return content_values["ASPSOL"]
+    
+    error_out("Unknown CONTENT keywords found in the aspect solution files")
+    
+
+
+
 def _verify_have_list(myfile,mylist,mydir,params):
     '''Verify we have list of fits files in output directory'''
 
@@ -184,6 +241,8 @@ def _verify_have_list(myfile,mylist,mydir,params):
 
         # sort files
         asol_files.sort()
+
+        asol_files = filter_aspsolobi_from_aspsol(asol_files,mydir)
 
         # write out sorted list
         try:


DougBurke:
The first warning in this case is a bit odd to me, since the file pointed to by ASOLFILE actually exists (as shown by the second warning), but the message makes it sound like it doesn't.

kglotfelty:
The first warning isn’t part of this change. It’ll actually happen for any obsid with multiple asol files (it’s checking file by file in the primary dir rather than on the stack of asols after all the contents of primary has been processed.)

bin/chandra_repro
```diff
@@ -137,6 +137,63 @@ def error_out(params,msg,val):
...
+    So if we find both ASPSOLOBI and ASPSOL files, we are going to pick
+    the ASPSOLOBI files.  This provides a better off-axis angle value
+    and tighter FOV.
+    """
+    from pycrates import read_file
+
+    content_values = {}
+
+    for f in asol_list:
+        tab = read_file(f.strip("\n").strip(" "), mode="r")
```
DougBurke, line 161: Really minor question: can you not just say `f.strip()` instead of the double `strip`?

bin/chandra_repro
```diff
@@ -137,6 +137,63 @@ def error_out(params,msg,val):
...
+    We know that there can be 1 and only 1 ASPSOLOBI file per obi (multi-obi
+    obsid's will need to have already been split, #repro5). 
+    
+    So if we find both ASPSOLOBI and ASPSOL files, we are going to pick
+    the ASPSOLOBI files.  This provides a better off-axis angle value
+    and tighter FOV.
+    """
+    from pycrates import read_file
+
+    content_values = {}
```
DougBurke, line 158: You could make this a dictionary with a default, which makes the addition bit simpler below. So at the top

```python
from collections import defaultdict
```

then here

```python
content_values = defaultdict(list)
```

so that later on you can just say

```python
content_values[content].append(f)
```

without having to check if `content` already exists

bin/chandra_repro
```diff
@@ -137,6 +137,63 @@ def error_out(params,msg,val):
...
+            error_out("Too many ASPSOLOBI aspect file found.  If this "+
+            "is a mutli-obi observations, be sure to split it first "+
+            "using the splitobs script")
+        else:
+            return content_values["ASPSOLOBI"]
+    elif "ASPSOL" in content_values:
+        return content_values["ASPSOL"]
+    else:
+        error_out("Unknown CONTENT keywords found in the aspect solution files")
+
```
DougBurke, line 193: So although the above `if` loop is how I'd write it, the Python linter I'm using really pushes you to change

```
if cond:
    <error out>
else:
    <blah>
```

to avoid the `else`, that is:

```
if  cond:
    <error out>

<blah>
```

and similarly for `return` statements. Which would suggest it would want this to be

```python
if "ASPSOLOBI" in content_values:
    if len(blah) > 1:
        error_out("xxx")

    return blahblah

if "ASPSOL in blah:
    return blahblah

error_out("xxx")
```

for appropriate polymorphic replacements for `xxx`, `blah`, and `blahblah`.

I can actually see a reason for this (thanks to Python's identation), even though it annoys my inner functional programmer.

DougBurke:
Note that none of my questions relate to the behavior of the code, it's just a matter of style.

bin/chandra_repro
```diff
@@ -137,6 +137,63 @@ def error_out(params,msg,val):
...
+        # But I do want to go ahead and remove the old files from the
+        # output directory.
+        for oldasp in content_values["ASPSOL"]:
+            old = os.path.join(mydir, 
+                      os.path.basename(oldasp.strip("\n").strip(" ")))
+            if os.path.exists(old):
+                os.unlink(old)
+
+    if "ASPSOLOBI" in content_values:
+        if len( content_values["ASPSOLOBI"] ) > 1:
```
DougBurke, line 183: Here's some "proper" questions (that really relate to different parts of the code than here...)

a) I think this should this check be moved to before the check on line 168 - i.e. immediately error out

b) I'm wondering if it's worth warning message on line 169 - how about if we just always use ASPSOLOBI in preference and don't bother telling the user in this case (it would mean we don't have to explain what the message means to people, and if a user really wants to be using the ASPSOL rather than ASPSOLOBI file I think it's okay for us to say they will have to do things manually / have the directory set up correctly [as we don't expect this to be a common occurrence]).

bin/chandra_repro
kglotfelty: Thanks.  Done.

bin/chandra_repro
kglotfelty: Yes. Done.

bin/chandra_repro
kglotfelty: My 1st commit was like that; but it was really hard to read (even the next day).  Especially w/ the multiline string.

But now that I've reworked the logic the "extra" else's have been purged.

bin/chandra_repro
kglotfelty: Okay, I've moved the `>1` check earlier.

I've kept the warning but up'ed it to `verbose=2`.

bin/chandra_repro
```diff
@@ -137,6 +137,66 @@ def error_out(params,msg,val):
...
+          "{} will be used".format(content_values["ASPSOLOBI"]))
+
+        # Wwant to go ahead and remove the old files from the
+        # output directory.
+        for oldasp in content_values["ASPSOL"]:
+            old = os.path.join(mydir, os.path.basename(oldasp.strip()))
+            if os.path.exists(old):
+                os.unlink(old)
+
+        # may not be necessary to del but I don't like relying on
```
DougBurke, line 183: An alternative would be to set up booleans earlier, like

```python
has_aspsolobi = 'ASPSOLOBI' in content_values
has_aspsol = 'ASPSOL' in concent_values
```

and then use these in the checks.

DougBurke:
LGTM - do you want it merged now or is there any more doc work needed (I don't think so, but wanted to check)?

kglotfelty:
hang on -- I need to check something

PR #371, opened 2020-04-03
HRC updated SAMP RMF

nplee:
this change uses calquiz to pick the appropriate RMF file for HRC data from the CALDB rather than using hardcoded paths


bin/specextract
```diff
@@ -947,23 +947,36 @@ def create_hrc_resp(specfile,rmf_file,refcoord,full_outroot,asp_param,clobber,ve
...
+                rmf = rmf[:rmf.find("[")]
+
             else:
                 raise ValueError("{0} has an invalid HRC detnam value.".format(specfile))
 
         elif chan == 256:
             # use non-SAMP responses
+            raise IOError("{0} is an invalid HRC spectral file, as non-SAMP responses are no longer supported by the CALDB.  Please reprocess the dataset or re-download the observation from the archive.".format(specfile))
+
             if detector.upper() == "HRC-S":
```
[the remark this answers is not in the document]
nplee: I've cleaned up the function removing a lot of these extraneous tests.

bin/specextract
```diff
@@ -937,56 +937,31 @@ def check_event_stats(file,refcoord,weights_check):
...
+        calquiz.calfile = "CALDB"
+        # calquiz.infile = "none"
+        # calquiz.telescope = "chandra"
+        # calquiz.instrument = "hrc"
+        # calquiz.calfile = "CALDB(DETNAM={})".format(get_keyval(specfile,"detnam"))
+
+        with suppress_stdout_stderr():
+            calquiz()
+
+        if calquiz.outfile == None:
```
DougBurke, line 955: @nplee - this should be `if calquiz.outfile is None` as there's a difference - e.g. http://jaredgrubb.blogspot.com/2009/04/python-is-none-vs-none.html

It shouldn't matter here, but I think it's a good habit to get into.

bin/specextract
```diff
@@ -937,56 +937,31 @@ def check_event_stats(file,refcoord,weights_check):
...
+            raise IOError("{0} is an invalid HRC spectral file, as non-SAMP responses are no longer supported by the CALDB.  Please reprocess the dataset or re-download the observation from the archive.".format(specfile))
 
         else:
```
DougBurke, line 958: Something my python liter is complaining about in my code is the use of

    if condition:
        raise SomeError()
    else
        do_something()

since it wants you to avoid the "redundant" `else` and write

    if condition:
        raise SomeError()
    
    do_something()

I think that makes sense to do here.

DougBurke:
@nplee - I think it's worth adding a change entry in the specextract ahelp file here - something basic like "A CALDB query is now used to select the HRC RMF file, rather than using a fixed value."

nplee:
@DougBurke I've updated the ahelp and modified the script with your conditional suggestions.

bin/specextract
```diff
@@ -937,56 +937,29 @@ def check_event_stats(file,refcoord,weights_check):
...
-            raise IOError("{0} is an invalid HRC spectral file".format(specfile))
+        with suppress_stdout_stderr():
+            calquiz()
 
-        # if specfile fails for obsfile, introduce use of event file, which adds another variable to this function
+        if calquiz.outfile is None:
+            raise IOError("{0} is an invalid HRC spectral file, as non-SAMP responses are no longer supported by the CALDB.  Please reprocess the dataset or re-download the observation from the archive.".format(specfile))
+
+        rmf = calquiz.outfile
+        rmf = rmf[:rmf.find("[")]
```
DougBurke, line 959: @kglotfelty - do you know if CALDB queries are guaranteed to end with a block specifier, or should we re-write this to something like the following (untested code) for safety (I'd rather not as it's just more to read)?

``` python
rmf = calquiz.outfile
lpos = rmf.find('[')
if lpos > -1:
    rmf = rmf[:lpos]
```

kglotfelty: Yes, it should always have a block number/name.  I haven't looked beyond this snipet, but as per your case on Friday, there can, in general, be multiple returns, which I believe should be comma separated.  [But generally if you've setup a specific enough query then we try to have only 1.]

DougBurke: So, there is the issue here that we could have multiple files in the output, so we should treat calquiz.outfile as a stack and then check on the number of files. That is

```python

# somewhere
import stk

rmfs = stk.build(calquiz.outfile)
if len(rmfs) > 1:
    # this should not happen, so do not spend a lot of time on the error message
    raise IOError("Multiple HRC RMFs returned by CALDB: {}".format(rmfs))

if len(rmfs) == 0:
    # this should not happen
    raise IOError("Unable to find HRC RMF")

rmf = rmfs[0]
rmf = rmf[:rmf.find("[")]
```

nplee: If we're going to cast the calquiz output as a stack, I think using the caldb4.Caldb module in lieu of calquiz will be cleaner.  The only reason I went with calquiz over caldb4 was to avoid playing with a list output.

PR #372, opened 2020-04-07
update combine_spectra to work around addresp NUM keywords bug

kglotfelty:
`addresp` is just copying the header from the 1st RMF file and not updating the NUMGRP and NUMELT keywords (should be the sum of the N_GRP and N_CHAN columns, respectively).  These are optional keywords, but if they are there, they should be correct.  Xspec 12.11.0 seems to use these and has been seen to crash when the values are incorrect.

Note: grating RMFs don't have those keywords, so if they are not in the input file, then this update will not create them in the output file.

### DIFF
--- a/bin/combine_spectra
+++ b/bin/combine_spectra
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #
-# Copyright (C) 2013,2014,2016-2019
+# Copyright (C) 2013,2014,2016-2020
 #       Smithsonian Astrophysical Observatory
 #
 # This program is free software; you can redistribute it and/or modify
@@ -21,29 +21,10 @@ from __future__ import print_function
 
 
 toolname = "combine_spectra"
-__revision__ = "18 November 2019"
+__revision__ = "07 April 2020"
 
 import sys
-
-
-# Make sure this shows up first!
 import os
-try:
-    if not __file__.startswith(os.environ['ASCDS_INSTALL']):
-        _thisdir = os.path.dirname(__file__)
-        _libname = "python{}.{}".format(sys.version_info.major,
-                                        sys.version_info.minor)
-        _pathdir = os.path.normpath(os.path.join(_thisdir, '../lib', _libname, 'site-packages'))
-        if os.path.isdir(_pathdir):
-            os.sys.path.insert(0, _pathdir)
-        else:
-            print("*** WARNING: no {}".format(_pathdir))
-
-        del _libname
-        del _pathdir
-        del _thisdir
-except KeyError:
-    raise IOError('Unable to find ASCDS_INSTALL environment variable.\nHas CIAO been started?')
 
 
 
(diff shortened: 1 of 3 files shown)


bin/combine_spectra
```diff
@@ -415,6 +396,53 @@ def fix_arf_exposure( arfs, outarf, method ):
...
+    numelt = sum( [sum(c) for c in chan]) # variable length array col
+    verb2("The updated NUMELT keyword is {}".format(numelt))
+
+    grp = tab.get_column("n_grp").values
+    numgrp = sum(grp)
+    verb2("The updated NUMGRP keyword is {}".format(numgrp))
+    
+    from ciao_contrib.runtool import dmhedit
+    dmhedit.infile=outrmf
+    dmhedit.file=""
```
DougBurke, line 428: We *could* avoid calling `dmhedit` twice and use the `filelist` option here (simplify the `dmhistory` output slightly, but at the expense of losing information on what has changed). I thought I had a routine to simplify this (i.e. generating the temporary file), but it turns out I don't. So I don't think it's a worthwhile change. But we now have proof that I decided against suggesting this!

DougBurke:
I guess the other option would be to delete the `NUMGRP` and `NUMELT` keywords if found?

bin/combine_spectra
```diff
@@ -415,6 +396,53 @@ def fix_arf_exposure( arfs, outarf, method ):
...
+    dmhedit.op="add"
+    dmhedit.datatype="ulong"
+    dmhedit.unit=""
+
+    if tab.get_key("NUMGRP") is not None:
```
DougBurke, line 433: How about checking that the keyword values are wrong before changing it (this future proofs the script a bit for when `addresp` is fixed)?

DougBurke:
Thinking about this a bit: I think we should move the update logic into a separate script, and then call it here. The reason being is that users may want to fix RMF files created with an old run of `combine_spectra` and don't want to have to re-run everything (or may not have the inputs).

I tested out pycrates `CrateDataset` class and it seems to do the right thing here (there's a change in the history number value, but not obviously anything else, at least not according to `dmdiff`):

```python
import os
import shutil

import pycrates

orig = 'aciss_9_src.rmf'
copy = 'copy.rmf'
# ds = pycrates.CrateDataset(orig, mode='r')

shutil.copy(orig, copy)
ds = pycrates.CrateDataset(copy, mode='rw')

matrix = ds.get_crate('MATRIX')

# should really use a case-insensitive lookup
n_grp = matrix.N_GRP.values
n_chan = matrix.N_CHAN.values

changed = False

numgrp = n_grp.sum()
k = matrix.get_key('NUMGRP')
if k is not None:
    k.value = numgrp
    # print("NUMGRP = {}  -> {}".format(k.value, numgrp))
    changed = True

# N_CHAN is a variable-length array which gets
# encoded as a Numpy vector of Numpy vectors.
#
n_chans = [sum(nc) for nc in n_chan]
numelt = sum(n_chans)

k = matrix.get_key('NUMELT')
if k is not None:
    k.value = numelt
    # print("NUMELT = {}  -> {}".format(k.value, numelt))
    changed = True

if not changed:
    os.remove(copy)
    sys.exit(0)

# Really should add a history record
print("Updated file: {}".format(copy))
ds.write(copy)
```

which gives

```
% dmdiff aciss_9_src.rmf copy.rmf
Infile 1:  aciss_9_src.rmf
Infile 2:  copy.rmf

-----------------------
HEADER VALUE DIFFERENCES
-----------------------


Message:                                Keyword:              Value(s):                    Diff:
--------                                 -------              ---------                    -----
Values are not equal                    CHECKSUM dZoJfWnJdWnJdWnJ  fa2LiS2IfY2IfY2I
# dmdiff (CIAO 4.12): WARNING: CHECKSUM comments are different.
# dmdiff (CIAO 4.12):   comment1= "HDU checksum updated 2019-03-18T04:24:15"
# dmdiff (CIAO 4.12):   comment2= "HDU checksum updated 2020-04-07T16:53:01"
Values are not equal                     HISTNUM               92  40                        -52
Values are not equal                      NUMGRP             1364  1382                      +18
Values are not equal                      NUMELT           382759  382017                   -742

--------------------------
SUBSPACE VALUE DIFFERENCES
--------------------------

Message:                         Column:       Row              Value(s):                    Diff:
--------                         -------       ---              ---------                    -----



TABLE NAME: MATRIX

-----------------------
TABLE VALUE DIFFERENCES
-----------------------

Message:                             Row:Cell:   Column:              Value(s):
     Diff:
--------                             ---------   -------              ---------
     -----

```

and

```
% dmdiff aciss_9_src.rmf"[ebounds]"
copy.rmf"[ebounds]"
Infile 1:  aciss_9_src.rmf[ebounds]
Infile 2:  copy.rmf[ebounds]

-----------------------
HEADER VALUE DIFFERENCES
-----------------------


Message:                                Keyword:              Value(s):                    Diff:
--------                                 -------              ---------                    -----


--------------------------
SUBSPACE VALUE DIFFERENCES
--------------------------

Message:                         Column:       Row              Value(s):                    Diff:
--------                         -------       ---              ---------                    -----



TABLE NAME: EBOUNDS

-----------------------
TABLE VALUE DIFFERENCES
-----------------------

Message:                             Row:        Column:              Value(s):
     Diff:
--------                             ----        -------              ---------
     -----

```

DougBurke:
Hmmm. With the "fixed" RMF (generated by my code) I seem to have problems reading things into Sherpa:

```
sherpa In [1]: load_pha('aciss_9_grp.pi')
read ARF file aciss_9_src.arf
WARNING: Index out of range
read background file aciss_9_bkg.pi
```

DougBurke:
I've tried fitting the data in XSPEC 12.11.0 with the correct NUMELT/NUMGRP keyword values and with these keywords deleted, and get the same result, so I think it's safe to just drop these ("optional but recommended" keywords for now).

kglotfelty:
I also could not find any mention of NUMGRP and NUMELT outside of mkacisrmf (aka calcrmf) and mkrmf (via rmflib).

I updated combine_spectra to simply delete those keywords if they exist.

PR #392, opened 2020-07-01
Added changelog for blanksky #326

DougBurke:


DougBurke:
@nplee - can you check I got this right: blanksky_sample doesn't have a method parameter, but it now supports this method, so I felt it easier to describe the changes this way, but maybe it's better described differently

nplee:
@DougBurke this description looks fine.  The update to blanksky should be transparent to users of blanksky_sample, since the change in blanksky_sample just ensures that the randomized value assigned to an event is <=1.

Issue #398, opened 2020-08-26
lightcurve clean report mean and sigma


kglotfelty:
Base on HD ticket 22909 , w/ obsid 17542 

using `deflare` with `clean`, it raises this exception

https://github.com/cxcsds/ciao-contrib/blob/27348431ec4d5520b9eec2cd35a249f7e43952c5/lightcurves.py#L950

it would be good if we included the mean and sigma that were computed here so users get a sense of what values are and why this exception is thrown.

DougBurke:
Closed by #401 (I assume)

PR #401, opened 2020-09-01
provide unclipped mean and sigma count rates from data when sigma clipping fails

nplee:
provide unclipped mean and n*sigma count rates from data when sigma clipping fails as requested by #398


DougBurke:
@nplee - can you add a release note about this change (I'd suggest rebasing since I've just merged the dax update to avoid a conflict).

nplee:
@DougBurke done, and I see you just merged and closed

PR #417, opened 2020-11-04
Remove update range

DougBurke:
Remove the update_column_range script and code to use it when reprojecting and merging event files. The functionality has not been needed since CIAO 4.6 (if not earlier).

One minor difference is that `update_column_range` could clip the TLMIN/MAX field to a .5 value, whereas we now just get the actual limit.

The runtool module has been updated to remove `update_column_range` and it has also picked up some other recent changes, both to the contrib code (eg #396) and CIAO


DougBurke:
@kjg - I don't see any reason not to do this, but thought I'd check (this script was written to address the old reproject_events behavior where the TLMIN/MAX wouldn't be updated, but that's no-longer been true for a while now).

kjg:
I'm guessing this is intended for @kglotfelty?

DougBurke:
Oops - sorry @kjg

kglotfelty:
I think the clip part is kinda important.  I get worried about the grids between events and exposure maps not matching up (since I think one starts at pixel center and the other at pixel edge -- going from memory on this).  I just don't know the effect.

DougBurke:
For the `fluximage` like tasks I spent some time trying to make sure we always use "sensible" ranges for  binning and image filtering (by extending to the outermost half-bin boundary on a scale that starts at 0.5 and uses the requested binning factor). So it should only be an issue for downstream tooling. That's not to say that's not important.

Because of this I hope the different tools are using the same scheme!

kglotfelty:
I was concerned about things like `reproject_obs` too , eg if you just `dmcopy` or even just load event file in `ds9`.  But it looks like `reproject_obs` is also setting the range to nice half pixel boundaries.

DougBurke:
Was this with this PR? I'd need to check reproject_obs to see if it
manually changes things, since I think a manual reproject_events used the
actual limits, rather than "nice" ones.

kglotfelty:
Yes, with this pr, but it’d be great to double check and make sure I didn’t foul it up.

DougBurke:
So, I think I'm probably not going to go with this at the moment. It does improve the code base, but only slightly, and I haven't done enough analysis to really know what the downstream consequences are.

Issue #400, opened 2020-08-27
specextract py3.8 literal comparison test


kglotfelty:
https://github.com/cxcsds/ciao-contrib/blob/27348431ec4d5520b9eec2cd35a249f7e43952c5/bin/specextract#L2155-L2156

In python 3.8, these `is not True` are throwing a warning

    INSTALL/bin/specextract:0: SyntaxWarning: "is not" with a literal. Did you mean "!="?
    INSTALL/bin/specextract:0: SyntaxWarning: "is not" with a literal. Did you mean "!="?

Looking at the lines, I don't think they logic is needed at all, and it should just be 

```python
blanksky = [kw for kw in blanksky if kw != "-99"]
mm_blanksky = [kw for kw in mm_blanksky if kw != "-99"] 
```

:question:

DougBurke:
Yeah, that looks like a lot of work for no gain!

PR #418, opened 2020-11-05
specextract logic clean up

nplee:
This update includes the logic clean up needed to eliminate many of the IDLisms in specextract.  This is an intermediary point in the script rewrite where some new functionality is disabled and/or commented out in the script.  Off-ramping at this point shows no differences in regression testing with prior versions of the script and will immediately address #400 although error message formatting has been changed.

### DIFF
--- a/bin/specextract
+++ b/bin/specextract
@@ -26,14 +26,15 @@ Script:
 
   specextract
 
-This script is an enhanced Python translation of the original CIAO tool specextract, which was written in S-Lang. It is run from the Unix command line, within the CIAO environment.
-
+	This script is an enhanced Python translation of the original CIAO tool
+	specextract, which was written in S-Lang. It is run from the Unix
+	command line, within the CIAO environment.
 """
 
 __version__ = "CIAO 4.12"
 
 toolname = "specextract"
-__revision__ = "06 April 2020"
+__revision__ = "02 November 2020"
 
 
 ##########################################################################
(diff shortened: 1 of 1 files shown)


bin/specextract
```diff
@@ -609,9 +463,9 @@ def set_badpix(evtfile, bpixfile, instrument, verbose):
 
         finally:
             if os.path.isfile(bpixfile) and getattr(ardlib,bpixpar.replace("-","_")).rstrip("[BADPIX]") == bpixfile:
```
DougBurke, line 465: I think this is better written as

    if not os.path.isfile(bpixfile) or getattr(...) != bpixfile:
        raise IOError(...)

as it is a bit clearer what leads to an error.

bin/specextract
```diff
@@ -798,28 +652,113 @@ def group_by_obsid(file_stack, file_type):
...
+              be explicitly purged and likewise if they are covered by a 
+              large flux, then they will be preserved and the ARF 
+              appropriately weighted.
+    """
+
+    cr = pcr.read_file("{}[wmap]".format(wmapfile))
+    im = cr.get_image().values
+    threshold = float(threshold)
+
+    cr.__del__()
```
DougBurke, line 706: I'm not sure what deleting the crate does to the values - I would suggest either

    im = cr.get_image().values.copy()

or just remove the `cr.__del__()` call (I prefer the second option).

bin/specextract
```diff
@@ -798,28 +652,113 @@ def group_by_obsid(file_stack, file_type):
...
+    dmimgthresh.cut = cutoff
+    dmimgthresh.value = "0.0"
+    dmimgthresh.expfile = ""
+    dmimgthresh.clobber = "yes"
+    dmimgthresh.verbose = "0"
+
+    dmimgthresh()
+
+    threshfile.close()
+    del(im)
```
DougBurke, line 757: If you really want, add the `del(cr)` here ;-)

bin/specextract
```diff
@@ -1122,10 +1057,10 @@ def event_stats(file, colname):
 
     elif str(colname) in ["x","y","chipx","chipy"]:
```
DougBurke, line 1058: The `str()` isn't helping us here so just say

    if colname in [...]

bin/specextract
```diff
@@ -1135,18 +1070,21 @@ def event_stats(file, colname):
...
 
-        elif colname=="chipx":
+        elif colname == "chipx":
             return int(float(src_x))
 
-        elif colname=="chipy":
+        elif colname == "chipy":
             return int(float(src_y))
 
+    cr.__del__()
```
DougBurke, line 1085: We should not need this - there was some code way back in CIAO 4.<very-small-number> which did, but if you have evidence we still need this then we need a cratedev bug report!

DougBurke:
@nplee - feel free when writing strings to use the f-string support added in Python 3.6 - ie change

    "foo={}  bar={}".format(foo, bar)

to

    f"foo={foo} bar={bar}"

It's not worth changing code that already exists, but as you come across new code (or edit old code) don't forget this trick!

nplee:
@DougBurke updated as we had discussed with your comments.

DougBurke:
Cheers

PR #420, opened 2020-11-05
Tweak warning message from merging event files

DougBurke:
For `merge_obs` and `reproject_obs`:

The warning message no-longer display the full difference, instead limiting to the number of decimal places in the limit (this means that values right at the limit will look to equal the limit but let's not worry about that as these limits are in some sense "made up" anyway).

We now also restrict the "accuracy" of the calculated tangent point (only for screen output, which @kglotfelty asked me about a **long** time ago). There's an interesting tweak there which leads to 36.92400... appear to be displayed as 36.83 but there's hysterical-raisons for this (the old version was trying to get "nice" output but doing it in an "interesting" way).

As an example:

# OLD

```
New tangent point: RA=7h 44m 16.703999999997734s Dec=37d 53' 36.924000000011574"
```

```
Warning: the merged event file x/merged_evt.fits
   should not be used to create ARF/RMF/exposure maps because
      the RA_NOM keyword varies by 0.002569232860011539 (limit is 0.0003)
      the DEC_NOM keyword varies by 0.030147682741002768 (limit is 0.0003)
      the ROLL_NOM keyword varies by 166.86669904322002 (limit is 1.0)
      the EXPTIME keyword contains: 0.4 3.1
        which means that the DTCOR value, and hence LIVETIME/EXPOSURE
        keywords are wrong
```

# NEW

```
New tangent point: RA=7h 44m 16.704s Dec=37d 53' 36.93"
```

```
Warning: the merged event file x/merged_evt.fits
   should not be used to create ARF/RMF/exposure maps because
      the RA_NOM keyword varies by 0.0026 (limit is 0.0003)
      the DEC_NOM keyword varies by 0.0301 (limit is 0.0003)
      the ROLL_NOM keyword varies by 166.9 (limit is 1.0)
      the EXPTIME keyword contains: 0.4 3.1
        which means that the DTCOR value, and hence LIVETIME/EXPOSURE
        keywords are wrong
```

### DIFF
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -13,6 +13,13 @@ Updated scripts
     These scripts should be more robust to future changes to the
     Chandra Data Archive.
 
+  merge_obs, reproject_obs
+
+    The warning to not use the merged event file for calculating the
+    instrument response no-longer displays the difference in all its
+    numerical glory, but instead limits to the number of decimal
+    places of the limit.
+
 Updated Python modules
 
   ciao_contrib.runtool
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -28,6 +28,12 @@ Updated Python modules
     The runtool module has been updated to reflect new tools and changes
     to the parameters of tools in the CIAO 4.13 release.
 
+  coords.format
+
+    The deg2ra() and deg2dec() functions have gained an optional ndp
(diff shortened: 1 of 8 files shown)


DougBurke:
@kglotfelty - how annoying will this be for regression tests?

kglotfelty:
minor annoying -- no worse than a new contam model update.

PR #426, opened 2020-11-16
chandra_repro use new asp_offaxis_corr tool to update asol files

### LINKED ISSUE #377: chandra repro recompute _AVG keywords
To review:  should chandra_repro recompute the DY_|DZ|DTH_AVG keywords to match the asol file?

Blaming Jean.

kglotfelty:
Currently a WIP , uses the new `asp_offaxis_corr` tool to combine and update asol files into aspsolobi file.

Compared results w/ pre-repro-5 data and repro-5 data for ACIS and HRC.  The problem is the actual asol itself has also changed (diff pcad calibrations), so exact diff is hard.  Grabbed the new (after repro5) asol file via chaser, renamed  "raw" columns back to original names and things looked spot on.

I'm guessing there will be discussion re: need to make separate obs.par (not using `dmmakepar` ).

Passing along to JCM for testing.

This will also close #377


bin/chandra_repro
```diff
@@ -211,6 +213,10 @@ def update_aspsol_to_aspsolobi(asol_list,mydir, params):
     from ciao_contrib.runtool import make_tool
     from pycrates import read_file
 
+    if params["asol_update"] is False:
+        v1("Skipping boresite update to aspect solution file")
+        return asol_list
+
     v1("Applying boresite update to aspect solution file")
```
kglotfelty, line 220: If the asol update is applied, then gratings users must set `recreate_tg_mask=yes`.  The sky coordinates in the region block of the evt2 file are no longer valid.

DougBurke:
Is this close to merging now or

- does it need more work
- does it need more testing
- does it need more coffee

kglotfelty:
It is close. I'd like someone other than me to run, but I doubt I'll get it.

DougBurke:
Well, my first randomly-selected obsid led to

```
Applying boresite update to aspect solution file
WARNING: The grating 0th order location must be recomputed when asol_update=yes.  Changing recreate_tg_mask=yes.
```

so I guess that's good

PR #437, opened 2020-12-07
specextract background response logic fixed

nplee:
cleaned up logic handling background responses and ANCRFILE/RESPFILE header keywords are appropriately set in background PI file.

### DIFF
--- a/bin/specextract
+++ b/bin/specextract
@@ -34,7 +34,7 @@ Script:
 __version__ = "CIAO 4.13"
 
 toolname = "specextract"
-__revision__ = "10 November 2020"
+__revision__ = "8 December 2020"
 
 
 ##########################################################################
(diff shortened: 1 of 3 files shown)


bin/specextract
```diff
@@ -3060,27 +3060,28 @@ def specextract(args):
                     
             if instrument == "ACIS":
                 try:
-                    if True in [weight == "yes", srcbkg == "bkg" and dobkgresp]: # force background to always be weighted
+                    if True in [srcbkg == "src" and weight == "yes", srcbkg == "bkg" and dobkgresp]: # force background to always be weighted
```
DougBurke, line 3063: It's better written

    if any([...]):

as `any` will return True if any of the elements in the input sequence are True. There's also an `all` function to require all elements of the sequence to be True, but that's not useful here.

bin/specextract
```diff
@@ -3060,27 +3060,28 @@ def specextract(args):
...
-                                raise IOError("Failed to PSF correct the ARF: {}".format(ancrfile))
+                                raise IOError("Failed to create ARF for {}".format(fullfile))
+
+                            if correct == "yes":
+                                v1("Calculating aperture correction for {0} ARF {1}".format(srcbkg,iteminfostr))
+                                
+                                try:
+                                    ancrfile = correct_arf(full_outroot, fullfile, filename, ancrfile, skyx, skyy, binarfcorr, clobber)
+                                except OSError:
+                                    raise IOError("Failed to PSF correct the ARF: {}".format(ancrfile))
```
DougBurke, line 3084: [Background reading, not a suggested code change]

For situations where we catch an error and then raise an error, I've started to occasionally use the

    raise BlahError(...) from None

syntax, since this cuts down on the huge chain of errors reported to the user. However, in this case we don't want that "optimisation" since it's useful to be able to find the actual error when handling helpdesk tickets and the like.

bin/specextract
nplee: @DougBurke these have been revised

DougBurke:
@nplee - the ahelp file needs updating. At least to change the date to December 2020, but worth a changes section just to highlight there have been code changes (even if we don't expect them to change any beavior)

nplee:
@DougBurke ahelp updated noting internal changes from #400 and #418, but otherwise should not have any expected change to the user.

DougBurke:
Ta

PR #446, opened 2021-01-11
Dax sherpa tweaks

### LINKED ISSUE #443: dax spectral fit enhancements
From @anetasie 

I was just running Sherpa fit using DAX. Is there an option for confidence on the fit parameters?
It would be nice to have it.

In the additional model options it would be good to have also absorption models which allows for additional intrinsic absorption component and in this case one does not need to check the colden, but use default settings, with redshift option which could be modified.
xszwabs, xszphabs, xsztbabs

kglotfelty:
This fixes #443 

- Add absorption models
- Add `conf` button (which required reorganizing the image fit logic)
- Fix flux estimates (set energy range)

### DIFF
--- a/share/doc/xml/dax.xml
+++ b/share/doc/xml/dax.xml
@@ -130,14 +130,18 @@ unix% ds9 -analysis $ASCDS_CONTRIB/config/ciao.ds9 ...
 </DESC>
 
 
-<ADESC title="Changes in the 4.13.1 (January 2021) release">
+<ADESC title="Changes in the 4.13.1 (March 2021) release">
   <PARA>
     Added additional redshifted absorption spectral models.
-    Added energy limits to the photon and energy flux calculations
+  </PARA>
+  <PARA>
+    Added energy limits to the photon and energy flux calculations.
+    Previously, the fluxes were computed for the entire energy range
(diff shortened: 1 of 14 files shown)


share/doc/xml/dax.xml
```diff
@@ -129,6 +129,18 @@ unix% ds9 -analysis $ASCDS_CONTRIB/config/ciao.ds9 ...
 
 </DESC>
 
+
+<ADESC title="Changes in the 4.13.1 (January 2021) release">
+  <PARA>
+    Added additional redshifted absorption spectral models.
+    Added energy limits to the photon and energy flux calculations
```
DougBurke, line 136: As mentioned in  https://github.com/cxcsds/ciao-contrib/issues/443#issuecomment-784319400 I think we should say that previously the flux was calculated over the full energy range of the instrument (so ~0.3-11 keV I think, but can't guarantee, so maybe we shouldn't give explicit values?)

DougBurke:
Thanks for the doc fixes. Unfortunately there's now a conflict in the changes file which could be "fixed" with a rebase. Or cause more fun.

kglotfelty:
foobar'ed by rebasing

Issue #209, opened 2019-03-01
should mkinstmap-calling routines set BPMASK=VFAINT when data is in this mode?


DougBurke:
Looking through http://cxc.harvard.edu/ciao/ahelp/mkinstmap.html#plist.detsubsys made me wonder if we should force `BPMASK=VFAINT` for `VFAINT`-mode data in the routines that create instrument maps (and presumably also ARFs).

kglotfelty:
The `VFAINT` vs. `FAINT` part is moot ... the `VFAINT` bits aren't set in `FAINT` mode anyways; and `VFAINT` is a superset of `FAINT`

What this would do (I THINK!) is to change the behavior of how afterglows are dealt with.  This is a longer discussion for a different thread.

DougBurke:
Would it affect streaks? Probably not.

kglotfelty:
No.

kglotfelty:
Excuse me while I :fork_and_knife: :chicken:  (git doesn't have a _crow_ emoji).

Digging into the `ardlib` code a bit for the recent helpdesk ticket -- you know the one -- it looks like I was confusing two different issues that directly relates to this.

Yes, we will want to include `BPMASK=VFAINT` (or probably `BPMASK=0x3ffff`) in the detsubsys calls for ACIS datasets.  All the other scripts which create responses (`specextract`, `mktgresp`, ... )

kglotfelty:
And to add another layer of confusion to all this ...

Yes, the code does recognize `VFAINT` and `FAINT` and sets the mask as described.

**BUT**

Way back in 2005 (circa ciao3.2.2) bits 9 & 10, the extra bits included in the `VFAINT` were obsoleted.  This happened at the same time we went from a 16bit to a 32bit status column. 

So indeed programmatically, there is a difference between `FAINT` and `VFAINT`, practically there is none as those bits are never set (and reading the old emails, should be actively unset if `acis_build_badpix` were to be rerun).  So that part is moot.

The other part about including the missing bit 17 (new shadow bit) is still relevant.

PR #450, opened 2021-01-25
Mktgresp shadow bpix

kglotfelty:
Following from #209 and Pat's helpdesk questions, `mktgresp` now adding `BPMASK=0x03ffff` to `detsubsys`.

### DIFF
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -28,6 +28,17 @@ Updated scripts
     
     Added button to run `conf` to the Model Editor GUI. 
 
+  mktgresp
+    The frame-store shadow is now included when calculating the
+    grating ARFs for ACIS observations. This means that there will
+    a change to the effective area at wavelengths which intersect the
+    bottom edge of the CCD (if any).  For more information see
+    https://cxc.cfa.harvard.edu/ciao4.13/caveats/acis_shadow_badpix.html
+
+    Updated to support HRC-I+LETG LSFPARM files if they are available
+    in the CALDB.  They are expected to be released in March 2021 with 
+    the Chandra CALDB 4.9.5 release.
+
 ## 4.13.0 - December 2020
 
 Updated scripts
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -1,14 +1,32 @@
 
-## 4.13.1 - January 2021
+## 4.13.1 - March 2020
 
-  mktgresp
+Updated scripts
+
+  download_obsid_caldb
   
-    Updated detsubsys parameter to recognize the new ACIS frame store 
-    shadow bit (17).
+    The script will now skip downloading CALDB files associated with
+    the transmission gratings (TG) if neither of the gratings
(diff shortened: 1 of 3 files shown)


bin/mktgresp
```diff
@@ -219,10 +219,38 @@ def make_tg_rmf( obsinfo, pha, spectrum, outroot, wvgrid_arf, wvgrid_chan):
...
+                products for HRC-I. An RMF will be created using 
+                LETG+HRC-S calibration products. This should be 
+                accurate for wavelengths < 60 A. For longer LETG+HRC-I 
+                wavelengths, the observed line widths can be increasingly 
+                broad compared with the RMF generated here owing to the 
+                flat HRC-I detector not following the geometry of the 
+                LETG Rowland torus."""
+                
+                msg = " ".join([x.strip() for x in msg.split("\n")])
+                verb0(msg)
```
DougBurke, line 256: Would it make sense to use the textwrap module here? https://docs.python.org/3.6/library/textwrap.html

Changes.CIAO_scripts
```diff
@@ -6,6 +6,9 @@
     Updated detsubsys parameter to recognize the new ACIS frame store 
     shadow bit (17).
 
+    Updated to support HRC-I+LETG LSFPARM files if they are available
+    in the CALDB.
```
DougBurke, line 10: Is there a CALDB version where they were added (so we can say something like "if you have CALDB vblah or later"), or is it more complicated taht that?

bin/mktgresp
kglotfelty: I took a look ; text wrap is overkill here (I'd end up setting width to 999 to put all on one line). Plus it's too picky about extra spaces.

Though oddly, the latest ver on this branch does not  have this, correct `.join()` it got lost along the way when I wrapped the other warning message.  So I'll  fix that.

Changes.CIAO_scripts
kglotfelty: There will be (assuming Dale's release happens at same time as contrib).

bin/mktgresp
DougBurke: One of the joys of asynchronous communication, GitHub-style, is that I don't know if the `join` issue has been fixed now?

DougBurke:
Unfortunately the changes file needs updating. Yay progress...

kglotfelty:
this is why I don't think we should update the Changes file until we actually are ready to build the r/c.  This is a waste of time.

kglotfelty:
I'll wait until the dax changes are merged and then I'll update for mktgresp.

DougBurke:
On the flip side

 - the changelog and ahelp edits helps understand the PR and any future PRs you may make
 - if you don't write it now, something will get forgotten

So I think the solution is that we don't add the date to the changelog and ahelp files until the release, using an easily-greppable symbol to indicate a change is needed.

kglotfelty:
at this point I have to start t his who pr over again (and now retest everything again) because it's also foobar'ed -- just because the changelog).

PR #460, opened 2021-02-25
specextract ARF BPMASK filtering

nplee:
added BPMASK filter affecting ARFs to account for framestore shadow; infrastructure updates and string formatting updated.

### DIFF
--- a/bin/specextract
+++ b/bin/specextract
@@ -34,7 +34,7 @@ Script:
 __version__ = "CIAO 4.13"
 
 toolname = "specextract"
-__revision__ = "8 December 2020"
+__revision__ = "25 Feb 2021"
 
 
 ##########################################################################
@@ -134,13 +134,13 @@ def check_files(stack, filetype):
     else:
         suffix = 's'
 
-    v2("Checking {0} file{1} for readability...".format(filetype, suffix))
+    v2(f"Checking {filetype} file{suffix} for readability...")
 
     for s in stack:
         # str() should not be needed, but left in, just in case
         filename = str(s)
-        v2("  validating file: {}".format(filename))
-        if filename in ["CALDB", "None", "None", "none"]:
+        v2(f"  validating file: {filename}")
+        if filename in ["CALDB", "None", "NONE", "none"]:
             v3("  ... skipping")
             continue
 
@@ -149,7 +149,7 @@ def check_files(stack, filetype):
             table = cxcdm.dmTableOpen(filename)
 
         except IOError:
-            raise IOError("{0} file {1} does not exist or could not be opened.".format(filetype,filename))
+            raise IOError(f"{filetype} file {filename} does not exist or could not be opened.")
 
         cxcdm.dmTableClose(table)
 
@@ -188,7 +188,7 @@ def check_file_pbkheader(infile_stack):
                 verify_keys.extend([headerkeys.has_key(kw) for kw in pbk_kw])
 
                 if False in verify_keys:
-                    v1("WARNING: {} missing header keywords.\n".format(fileio.remove_path(inf)))
+                    v1(f"WARNING: {fileio.remove_path(inf)} missing header keywords.\n")
 
                     file_status.append(inf)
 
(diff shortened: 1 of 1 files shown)


bin/specextract
```diff
@@ -134,13 +134,13 @@ def check_files(stack, filetype):
...
-    v2("Checking {0} file{1} for readability...".format(filetype, suffix))
+    v2(f"Checking {filetype} file{suffix} for readability...")
 
     for s in stack:
         # str() should not be needed, but left in, just in case
         filename = str(s)
-        v2("  validating file: {}".format(filename))
-        if filename in ["CALDB", "None", "None", "none"]:
+        v2(f"  validating file: {filename}")
+        if filename in ["CALDB", "None", "NONE", "none"]:
```
DougBurke, line 143: I strongly suggest

```
if filename.upper() in ["CALDB", "NONE"]:
```

bin/specextract
```diff
@@ -1326,8 +1374,8 @@ def get_filename(full_filename):
 
 
 
-    if not filename or str(filename) in ["NULL", "NONE", "None",""," "]:
-        return NULL
+    if not filename or str(filename) in [None, "NULL", "NONE", "None",""," "]:
```
DougBurke, line 1377: str(filename) can't be None.

bin/specextract
```diff
@@ -1326,8 +1374,8 @@ def get_filename(full_filename):
...
+        return None
 
     else:
```
DougBurke, line 1380: pylint (or pylakes or ....) suggests it's bad form to have logic like

```
if some_conditon:
    ... code
    return value1
elsr:
    ... other code
    return value2
```

since the claim is it's better to say

```
if some conditon:
    ... code
    return value1

... other code
return value2
```

bin/specextract
```diff
@@ -1412,8 +1460,8 @@ def get_region_filter(full_filename):
         region = full_filename
 
 
-    if not region or str(region) in ["NULL", "NONE", "None",""," "]:
-        raise IOError("Please specify a valid spatial region filter for {} or use FOV region files.".format(full_filename))
+    if not region or str(region) in [None, "NULL", "NONE", "None",""," "]:
```
DougBurke, line 1463: As you have this logic in several places I would suggest creating a function like

```
# note: not checked it matches the logic exactly
def check_filename_is_set(filename):
    if not filename:
        return False
    filename = filename.strip().upper()
    if filename in ["", "NULL", "NONE"]:
        return False
    return True
```

for use here.

bin/specextract
```diff
@@ -2370,60 +2264,55 @@ def specextract(args):
...
         dtf_count = len(dtf_stk)
+        
     # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-
+    #
     #
     # check that there are no spaces in ancillary file paths, if specified
     #
     for path in asp_stk:
```
DougBurke, line 2294: this could all be put into a loop as you did earlier, rather than having lots of repeated code.

bin/specextract
```diff
@@ -2741,7 +2625,7 @@ def specextract(args):
             if fcount == 1:
                 iteminfostr = "\n"
             else:
-                iteminfostr = "[{0} of {1}]\n".format(str(ii+1),str(fcount))
+                iteminfostr = f"[{str(ii+1)} of {str(fcount)}]\n"
```
DougBurke, line 2628: The `str` should not be needed here, I think

DougBurke:
We'll also need ahelp update ;-)

There also appeared to be a change in the WMAP handling (or is this code that is currently unused)?

nplee:
> We'll also need ahelp update ;-)
> 
> There also appeared to be a change in the WMAP handling (or is this code that is currently unused)?

No change in WMAP handling since the new snippet is currently unused but was made from the discussion from AF testing prism of which WMAP to have alongside the PI file.

Issue #453, opened 2021-02-16
chandra_repro make hrc-s+LETG PI filter optional


kglotfelty:
Per AF: add parameter to make optional.

PR #461, opened 2021-02-25
chandra_repro add _NOM to asol and pi_filter option

### LINKED ISSUE #444: chandra_repro needs to add _NOM to updated asol
The `asp_offaxis_corr` tool deletes the `NOM` keywords from the asol file header

https://github.com/cxcsds/ciao-contrib/blob/82de62479d5615ace53b47f459ccf94fb625f415/bin/chandra_repro#L267

but somewhere in SDP, those keywords are added back (probably the same place where the `CONTENT` keyword is modified).

`chandra_repro` needs to add the `NOM` keywords back, with values equal to the `PNT`s.

kglotfelty:
This fixes #444 and #453

I'll make a separate PR w/ the Changes after mktgresp has been merged.

-    Now updates the reprocessed aspect solution file, pcadf*_repro_asol1.fits,    to include the RA_NOM, DEC_NOM, and ROLL_NOM.  They keywords    are removed the by asp_offaxis_corr tool but are required by    MARX.

-    The new parameter, `pi_filter`, controls whether the HRC-S+LETG    background PI filter is applied or not


DougBurke:
We'll need to remember to re-generate runtool when the time comes.

PR #464, opened 2021-03-04
tweak for per-repro-4 (early) data

kglotfelty:
Early data (not through repro-4) does not have the `D*_AVG` keywords.  This isn't fatal, but if left unchecked will cause `chandra_repro` to print warnings like:

```bash
pset: parameter not found : dy_avg
```

### DIFF
--- a/bin/chandra_repro
+++ b/bin/chandra_repro
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 #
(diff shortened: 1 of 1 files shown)


DougBurke:
Would it make sense to add these values to the output file even for such old data, or does that way madness lie?

DougBurke:
okay

DougBurke:
I don't think this needs an update to the changelog.

PR #465, opened 2021-03-04
download_chandra_obsid no-longer downloads vvref2 files by default

### LINKED ISSUE #463: download_chandra_obsid differentiate vv and vvref files
As per today's helpdesk ticket, the vvref2.pdf is often the largest file for an obsid; yet it contains little of interest to most users beyond what is in the much smaller vv2.pdf file. 

Most users just want the comments -- I wish there was a public API for these.  But since not, it would be nice if download_chandra_obsid would treat these two files as different file types: `vv` and `vvref`

DougBurke:
download_chandra_obsid no only downloads the vvref2.pdf.gz file if explicitly asked for. The vvref option has been added to select these files, and the vv option now only selects the vv2.pdf file.

Fixes #463 but is a **tad** ugly.

### DIFF
--- a/ciao_contrib/cda/data.py
+++ b/ciao_contrib/cda/data.py
@@ -368,10 +368,18 @@ class ObsId:
 
     def filter_files(self, types=None, formats=None):
         """Filter the list of files by the given types and/or formats.
+
+        If types is None then we exclude the vvref file type; that is if
+        you want to download the vvref2.pdf file you need to supply
+        a list of types and include the ''vvref' option. This is not
+        a great design!
         """
 
         V3(f"Before filtering: {len(self.files)} files")
-        if types is not None:
+        if types is None:
+            # exclude the vvref file
+            self.files = [f for f in self.files if not f.is_type(['vvref'])]
+        else:
             self.files = [f for f in self.files if f.is_type(types)]
         if formats is not None:
             self.files = [f for f in self.files if f.is_format(formats)]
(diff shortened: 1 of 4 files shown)


DougBurke:
I need to add a changelog entry.

ciao_contrib/cda/data.py
```diff
@@ -340,42 +341,50 @@ def __init__(self, obsid, base_url, hdr):
...
+        a list of types and include the ''vvref' option. This is not
+        a great design!
         """
 
-        V3("Before filtering: {} files".format(len(self.files)))
-        if types is not None:
+        V3(f"Before filtering: {len(self.files)} files")
+        if types is None:
+            # exclude the vvref file
+            self.files = [f for f in self.files if not f.is_type(['vvref'])]
```
kglotfelty, line 381: Humm ... I'm not sure.  Might need others to comment.

If I don't specify any filetypes then I'm either expecting everything or just the most useful files. 

I can also easily argue that the secondary/aspect, secondary/ephem, and oif.fits are not generally useful.  I've never used the ephem/ files and only used the osol files once.  The aspect adat71 files needed for monitor star observations are in primary/ when they it was requested (otherwise they are in 'supporting' only available via chaser).  axbary and sso_freeze use the orbit ephem which is in the primary dir.   These are all small, but often a bunch of them.

Just don't want to start a turf war w/ archive as to what should/shouldn't be in the standard distribution.

If you leave out the condition with `if not f.is_type(['vvref'])` would that keep those files when no filetypes is specified and still let 'vv' just d/l the vv2.pdf file?

DougBurke: I believe what the user actually wants is that they just want to just not have the vvref2.pdf file downloaded without them having to make any change. If we want to do this then you have the trickiness of the last commit in some form (and then we can think about excluding  other file types by default).

So options include
 - some form of the current code
 - explicitly come up with a list of file types we want to download by default (perhaps with a label at the download "standard" and get a known set of files)
 - add a way to say "don't download these file types" and then tell the user to set this to vvref (which negates the "user does not need to think about this", but is conceptually cleaner)

The way the current "labelling" code works is that we should have a label for all file types, so I don't understand your last point.

DougBurke: I think I'm going to go back to 

1. `download_chandra_obsid obsid` downloads everything (so the helpdesk user will need to jump to point 2)
2. add a `--exclude` flag (name TBD) so you can `download_chandra_obsid --exclude vvref obsid`to exclude the vvfref2 file
3. come up with some decision if you include and exclude a particular file type...

kglotfelty: :ok:

Note: i'm off today so I won't be able to review anything till Monday.

ciao_contrib/cda/data.py
```diff
@@ -458,29 +471,50 @@ def get_http_header():
...
+        which files to download. See the known_file_types array in
+        this module for the list of supported types.
+    excludes : sequence of str or None, optional
+        Those file types to ignore. It is expected that either (or
+        both) filetypes or excludes is None. If they are both given
+        then the filetypes setting "wins out" if a filetype is listed
+        in both.
+    mirror : str or None, optional
+        If None then use the CDA HTTPS archive is used (the BASE_URL
+        settign), otherwise set it to the name of a mirror archive to
```
kglotfelty, line 499: typo 'setting'

kglotfelty (approved):

PR #467, opened 2021-03-10
fix issue with ciao-install editions

### LINKED ISSUE #442: check_ciao_version
https://github.com/cxcsds/ciao-contrib/blob/82de62479d5615ace53b47f459ccf94fb625f415/bin/check_ciao_version#L191

`ciao` is undefined.    Fails with the `ciao-install` editions.

(I'm guessing then that this isn't called in the `conda` edition?)

kglotfelty:
This fixes #442

### DIFF
--- a/bin/check_ciao_version
+++ b/bin/check_ciao_version
@@ -2,7 +2,7 @@
 
 """
 
-Copyright (C) 2011, 2014, 2015, 2016, 2019, 2020
+Copyright (C) 2011, 2014, 2015, 2016, 2019, 2020, 2021
           Smithsonian Astrophysical Observatory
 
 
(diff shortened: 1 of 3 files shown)


bin/check_ciao_version
```diff
@@ -188,7 +188,7 @@ def compare_caldb():
         cinfo += f" (link to {lname})"
 
     if rval is None:
-        print(f"{ciao} is up to date.")
+        print(f"{caldb} is up to date.")
```
DougBurke, line 191: I think I probably meant it to be

```
print(f"{cinfo} is up to date.")
```

but I'm not sure how bothered I am about it now.

PR #474, opened 2021-05-26
specextract bug fixes

nplee:
minor bug fixes on top of the specextract released in contrib 4.13.1 quashing misleading 'error' and informational messages due to a try/except block when dealing with blanksky background files and a bug that causes combining results to fail even when 'combine=yes'.

### DIFF
--- a/bin/specextract
+++ b/bin/specextract
@@ -34,7 +34,7 @@ Script:
 __version__ = "CIAO 4.13"
 
 toolname = "specextract"
-__revision__ = "09 Mar 2021"
+__revision__ = "26 May 2021"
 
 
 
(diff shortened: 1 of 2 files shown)


DougBurke:
Can you add a changes section to the ahelp file?

nplee:
@DougBurke I've updated the ahelp file under the same commit.

bin/specextract
```diff
@@ -2911,15 +2921,18 @@ def specextract(args):
...
 
-                    elif "bkg" == srcbkg:
-                        cs_bkg_spectra[ii] = 1
-                        bphafiles[ii] = phafile
+            if docombine:
+                if srcbkg == "src":
+                    cs_src_spectra[ii] = 1
+                    sphafiles[ii] = phafile
 
+                elif srcbkg == "bkg":
```
DougBurke, line 2931: A questino for my amusement: can `srcbkg` be anything other than "src" or "bkg"?

nplee: no; in the infrastructure updates I'm making, I'm trying to get rid of these "elif" calls when the choice is binary.

DougBurke: Ta

PR #477, opened 2021-05-27
install_marx: add symlink to versioned dir

### LINKED ISSUE #471: install_marx add symlink
JCM:

> If we put a soft link in `$ASCDS_INSTALL/marx -> marx-5.5.1`, then I could have an alias 

```bash
alias setmarx='source ${ASCDS_INSTALL}/marx/setup_marx.csh'
```
> that wouldn't need to be changed ever (i.e. when the marx version changes).

KJG: need to support updates (so del and re-link if new version is installed).

kglotfelty:
This fixes #471

### DIFF
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -1,12 +1,4 @@
 
-## 4.13.2 - June 2020
-
-  install_marx
-  
-    Create a "marx" symbolic link to the version specific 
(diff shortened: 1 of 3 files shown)


kglotfelty:
Note: I'll fix 2020 -> 2021 in ChangeLog with another PR when I add dax and simpsf changes.

Issue #482, opened 2021-06-25
Can we suggest an energy-band to handle contamination changes?


DougBurke:
One for @kglotfelty.

Given an observation, hence observation range (possibly just TSTART as long as we can ignore merged event files) can we come up with a suggested "soft-band" mid-point based on the contamination model? The idea would be to say "hey, perhaps use this rather than the CSC 's' band". 

Of course, this then leads to doing the same for n observatiions - do we want to come up with a single band definition (the question would be how to wait the infividual band mid-points), or group the observations into chunks each with a different band, or give up on that idea and make it up to the user.

kglotfelty:
well, there's

https://cxc.cfa.harvard.edu/ciao/why/low_energy_mono.html#mean_energy

which of course was generated using whatever the then current version of the contam model, and using the same spectral model that was used to come up with the 0.92 value; figure 5 shows  the effect of varying the model parameters.

To calculate this "on the fly" you need and ARF+RMF, a spectral model, and your choice of metric (ie just because we use mean, maybe the max is more appropriate).  The code to generate those plots is in

/data/lenin2/Projects/Bands/soft_band_v_ao.ipynb

and there's additional info in

/data/lenin2/Projects/Bands/SoftBand.pdf

DougBurke:
So, do we want to move this into `ciao-contrib` (it's not complex but might be nice to make available)?

PR #483, opened 2021-06-28
new tool to compute characteristic monochromatic energy

kglotfelty:
Based on discussion in  #482 .

Note: if you want an unweighted then just use `model=const1d.flat paramval=flat.c0=1`

### DIFF
--- a/bin/find_mono_energy
+++ b/bin/find_mono_energy
@@ -38,10 +38,16 @@ def setup_dataspace(arf, rmf):
 def setup_model(model, model_params):
     "Set the source model"
 
-    # pylint: disable=exec-used
+    try:
+        set_model(DATASET_ID, model)
+    except Exception as mybad:
+        raise ValueError(f"ERROR in model '{model}'\n{mybad}")
 
-    set_model(DATASET_ID, model)
-    exec(model_params)
+    try:
+        # pylint: disable=exec-used
+        exec(model_params)
+    except Exception as mybad:
+        raise ValueError(f'Error in paramvals={model_params}]\n{mybad}')
 
 
 def setup_band(band):
@@ -93,6 +99,7 @@ def clear_output():
     paramclose(pfile)
 
 
+@lw.handle_ciao_errors(TOOLNAME, __REVISION__)
 def main():
     "Main routine"
 
@@ -113,9 +120,9 @@ def main():
     setup_model(pars["model"], pars["paramvals"])
     setup_band(pars["band"])
     mono_energy = compute_metric(pars["metric"])
-    
+
     verb1(f"Characteristic Energy = {mono_energy}")
-    
+
     pset_ouput(mono_energy)
 
 
--- a/bin/find_mono_energy
+++ b/bin/find_mono_energy
@@ -206,14 +206,10 @@ def clear_output():
 def main():
     "Main routine"
 
-    # Pre-process the argument list to handle special case
-    #
-    args = lw.preprocess_arglist(sys.argv)
-
     clear_output()
 
     # Load parameters
-    pars = get_params(TOOLNAME, "rw", args,
+    pars = get_params(TOOLNAME, "rw", sys.argv,
                       verbose={"set": lw.set_verbosity, "cmd": verb1})
 
     setup_dataspace(pars["arffile"], pars["rmffile"])
--- a/bin/find_mono_energy
+++ b/bin/find_mono_energy
@@ -67,16 +67,34 @@ def setup_dataspace(arf, rmf):
 
 
 def setup_model(model, model_params):
-    "Set the source model"
+    """Set the source model
+
+    Parameters
+    ----------
+    model : str
+        The model expression containing modelname.instancename terms
+        following Sherpa.
+    model_params : str
+        The parameter expression. This can actually be any valid Python
+        string but is expected to be instancename.paramname=value terms
+        separated by semi-colons. It can also contain commands like
+        'set_xsabund("lodd")'. This string can be empty.
+
+    """
 
     try:
         ui.set_model(DATASET_ID, model)
     except Exception as mybad:
         raise ValueError(f"ERROR with model={model}\n{mybad}") from mybad
 
+    if model_params.strip() == '':
+        return
+
     try:
+        # Evaluate the parameter expressions within the ui namespace.
+        #
         # pylint: disable=exec-used
-        exec(model_params)
+        exec(model_params, vars(ui))
     except Exception as mybad:
         raise ValueError(f'Error with paramvals={model_params}\n{mybad}') from mybad
 
--- a/bin/find_mono_energy
+++ b/bin/find_mono_energy
@@ -38,7 +38,7 @@ from ciao_contrib.param_soaker import get_params
 
 
 TOOLNAME = "find_mono_energy"
-__REVISION__ = "10 November 2021"
+__REVISION__ = "12 November 2021"
 
 DATASET_ID = "myid"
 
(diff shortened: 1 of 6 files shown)


bin/find_mono_energy
```diff
@@ -0,0 +1,123 @@
...
+    set_rmf(DATASET_ID, _rmf)
+
+
+def setup_model(model, model_params):
+    "Set the source model"
+
+    # pylint: disable=exec-used
+
+    set_model(DATASET_ID, model)
+    exec(model_params)
```
DougBurke, line 44: Perhaps we should have this in a `try/except` block so we can say something like (but tested)

````
try:
    exec(model_params)
except Exception as e:
    raise ValueError(f'Error in paramvals={model_params}]\n{e}')
```

bin/find_mono_energy
```diff
@@ -0,0 +1,123 @@
...
+    csc = {'soft': '0.5:1.2', 'medium': '1.2:2.0', 'hard': '2.0:7.0',
+           'broad': '0.5:7.0', 'ultrasoft': '0.2:0.5',
+           'wide': '0.1:10.0'}
+
+    if band in csc:
+        elo, ehi = [float(x) for x in csc[band].split(":")]
+    elif band.count(":") != 1:
+        raise ValueError(f"Unknown band value {band}")
+    else:
+        elo, ehi = [float(x) for x in band.split(":")]
```
DougBurke, line 59: If a user sets band to `":7"` then this will raise a `ValueError` saying 

    could not convert string to float: ''

which could be a little confusing but I'm not sure it's worth wrapping this up to catch the error.

bin/find_mono_energy
```diff
@@ -0,0 +1,123 @@
...
+
+    notice_id(DATASET_ID, elo, ehi)
+
+
+def compute_metric(metric):
+    "Compute the metric based on the fluxed spectrum"
+    dater = get_model_plot(DATASET_ID)
+
+    emid = (dater.xlo+dater.xhi)/2.0
+    flux = dater.y
```
DougBurke, line 69: These are in counts/s/keV. I would have thought you'd want to remove the bin width (keV) for the max case at least (too early in the week for me to work out whether it makes a difference for the mean case).

Now, for chandra data we generally have equal-width bins so it shouldn't matter for the max case, but can we guarantee it?

param/find_mono_energy.par
```diff
@@ -0,0 +1,10 @@
+arffile,f,a,"",,,"Input ARF file"
+rmffile,f,a,"",,,"Input RMF file"
+model,s,a,"xswabs.abs1*xspowerlaw.pwrlaw",,,"Sherpa model expression"
```
DougBurke, line 3: The `xswabs` model is "old" and I wonder if we should use a more-modern equivalent (e.g. `xsphabs`). I don't expect it to make a significant difference as we are looking at significantly-larger scales than we see differences. For funsies

```
sherpa In [7]: xsphabs.phabs
-------------> xsphabs.phabs()
Out[7]: <XSphabs model instance 'xsphabs.phabs'>

sherpa In [8]: xswabs.wabs
-------------> xswabs.wabs()
Out[8]: <XSwabs model instance 'xswabs.wabs'>

sherpa In [9]: phabs.nh = 0.2

sherpa In [10]: wabs.nh = 0.2

sherpa In [11]: egrid = np.arange(0.1, 1, 0.01)

sherpa In [12]: yphabs = phabs(egrid[:-1], egrid[1:])

sherpa In [13]: ywabs = wabs(egrid[:-1], egrid[1:])

sherpa In [14]: emid = (egrid[:-1] + egrid[1:]) / 2

sherpa In [15]: plt.clf()

sherpa In [16]: plt.plot(emid, ywabs, label='wabs')
Out[16]: [<matplotlib.lines.Line2D at 0x7ff22022ee50>]

sherpa In [17]: plt.plot(emid, yphabs, label='phabs')
Out[17]: [<matplotlib.lines.Line2D at 0x7ff22036fe50>]

sherpa In [18]: plt.legend()
Out[18]: <matplotlib.legend.Legend at 0x7ff23001b670>

sherpa In [19]: plt.xlabel('Energy (keV)')
Out[19]: Text(0.5, 120.87222222222219, 'Energy (keV)')
```

x

DougBurke:
I'd actually have to look at the code, and that's obviously too hard for me...

You say

> Note: if you want an unweighted then just use model=const1d.flat paramval=flat.c0=1

Should it be `const1d.flat` or `scale1d.flat`? It only matters if the model being fit is PHA-like (ie has low and high energies).

kglotfelty:
No idea ... I never understood the integrate flag thing or if it matters here.  (not fitting anything)

DougBurke:
@kglotfelty - I've rebased this and made  a few minor tweaks. The calculation should be the same as before but I haven't actually checked it.

kglotfelty:
Humm, I'm also not sure why you need to `preprocess_args` thing .  The `--debug|--tracebackdebug` stuff should be working normally without it.  It's the same basic `main` that I use in all my stuff, and no problems using `--debug`.

DougBurke:
Well, it works for me.... Using --debug didn't seem to be working for me until I added the pre-processing but maybe I missed something.

kglotfelty:
I'll take a look on Friday.

DougBurke:
If I don't call `preprocess_arglist` then I can't use --debug or --tracebackdebug 

```
% find_mono_energy blob_src1.arf blob_src1.rmf band=soft model=const1d.flat paramv=flat.c0=100 metric=mean --tracebackdebug
Sorry, Arguments not ok, try again.
# find_mono_energy (10 November 2021): ERROR paramopen() Could not execute paramopen
% find_mono_energy blob_src1.arf blob_src1.rmf band=soft model=const1d.flat paramv=flat.c0=100 metric=mean --debug
Sorry, Arguments not ok, try again.
# find_mono_energy (10 November 2021): ERROR paramopen() Could not execute paramopen
```

DougBurke:
For the `ui.` change - I was actually worried about that, but it somehow magically works. I don't actually understand how, particularly given the following:

```
In [1]: xsphabs.gal
NameError                                 Traceback (most recent call last)
<ipython-input-1-d6e5e258cb5c> in <module>
----> 1 xsphabs.gal

NameError: name 'xsphabs' is not defined

In [2]: from sherpa.astro import ui

In [3]: xsphabs.gal
NameError                                 Traceback (most recent call last)
<ipython-input-3-d6e5e258cb5c> in <module>
----> 1 xsphabs.gal

NameError: name 'xsphabs' is not defined

In [4]: ui.xsphabs.gal
Out[4]: <XSphabs model instance 'xsphabs.gal'>

In [5]: eval('xsphabs.gal')
NameError                                 Traceback (most recent call last)
<ipython-input-5-9f920fba0d02> in <module>
----> 1 eval('xsphabs.gal')

<string> in <module>

NameError: name 'xsphabs' is not defined
```

kglotfelty:
Okay, w.r.t. the `preprocess` thing, you only need to use it in the `pset_output` routine, not in the `main` routine.

I'm remember I tried using the `ciao_contrib.param_wrapper.open_param_file` but it doesn't let you set the mode, and I needed to set the mode to include "H" to avoid getting prompted for parameters again when I re-opened it to save the value.

DougBurke:
I just added the probably-not-to-be-advertised-at-this-time functionality that lets you run with only one of the responses, to answer the age-old question "which would win. ARF or RMF. Fight!"

```
% find_mono_energy arf=blob_src1.arf rmf=blob_src1.rmf band=soft  mode=h v=0; pget find_mono_energy energy
0.8933711077874543
% find_mono_energy arf= rmf=blob_src1.rmf band=soft  mode=h v=0; pget find_mono_energy energy
0.7941796208701266
% find_mono_energy arf=blob_src1.arf rmf= band=soft  mode=h v=0; pget find_mono_energy energy
0.8895071142784986
```

DougBurke:
Aha - I now know why we can create models without the `ui.` prefix. The model expression is sent to `ui.set_source` which internally ends up calling `eval(expr, self._model_globals, self._model_components)` - so it an access these symbols without the `ui.` label.

PR #488, opened 2021-07-30
add ACIS, HRC, and TG filename keywords to merge rules

kglotfelty:
This adds the various `FILE` name keywords to the merge rules.

Note:

- I left the `ASOLFILE`, `MASKFILE`, and `BPIXFILE` alone since `merge_obs` and family does something special with those, and
- The `BIASFIL` (no number) catches all the `BIASFIL0` - `BIASFIL9` keywords.  Similarly the `LIMFILE` (from HRC) catches the `LIMFILE1` keyword.

I will be sending an RFE for `dmmerge` 's default lookup table to dsdevel shortly to make the same update.


DougBurke:
I don't think I'll take this for the September release - it's one of those "should not be a  problem but may just trigger something in some random script somewhere" types of update that I need to look at properly...

kglotfelty:
agreed

DougBurke:
Hmmm, there's some changes that have happened over time that mean merge_obs/flux_obs fall through to the fluximage rules rather than the specific one they are meant to use for. Oh joy.

data/obsidmerge_header_lookup.txt
```diff
@@ -54,3 +54,24 @@ TIMEUNIT   Fail;Default-s
 TIMESYS    Fail;Default-UTC
 TIMEPIXR   Fail;Default-0.5
 OBI_NUM    Skip
+THRFILE    Fail
```
DougBurke, line 57: Having these as Fail created a bunch of messages like the following (once I fix some things). I think it would make sense for the obsidmerge case to have these as `Skip` rather than `Fail` - does that make sense?

```Combining 4 observations.
BIASFIL2 values are different...FAIL...
BIASFIL3 values are different...FAIL...
BIASFIL5 not present in all input files...FAIL...
BIASFIL6 values are different...FAIL...
...
```

kglotfelty: Yup, `Skip` is probably a better option.  I remember I wasn't sure if it was supported for string data-types (`Fail` wasn't until recently-ish) but a quick check looks like it is.

PR #490, opened 2021-08-10
chandra_repro: update CONTENT and HDUCLAS2 keywords

### LINKED ISSUE #489: chandra_repro metadata
From Dave P:

The evt2 file keywords are just coming directly from the evt1 file.  We should really modify a couple of them to indicate an L2 file:
- `CONTENT` should be `EVT2`
- `HDUCLAS2` should be `ACCEPTED`
- probably more

and no, we don't need the `pyhdrlib` to do this :)

kglotfelty:
This fixes #489 

only affects keywords; evt2 is/has always had standard filters applied.

### DIFF
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -7,6 +7,11 @@ rather than a TTY.
 
 Updated scripts
 
+  chandra_repro
+
+    The reprocessed event file now has a CONTENT keyword set to "EVT2"
+    and HDUCLAS2 keyword set to "ACCEPTED".
+
   check_ciao_version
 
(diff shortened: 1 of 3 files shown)


DougBurke:
We may as well take this (it's small and I assume inoffensive) - or do you disagree @kglotfelty ?

I can add the changelog / update the release date.

kglotfelty:
I would leave this back.  If nothing else, it will break all our regression tests.

I'm guessing that at the least the ahelp will need to be rebased w/ the changes in #494 which given my github experiences is more than I really wanted to do right now.

DougBurke:
Okay - I'll leave.

PR #491, opened 2021-09-02
WIP: specextract revamp

nplee:
initial reviewable form of the revamped specextract, have passed standard regression test and seemingly functional.  
Front load most checks so error outs early on and setup for the runs in ciao_contrib/_tools/specextract.py


DougBurke:
@nplee -the first commit is (I'm guessing) the same as #496 - do we need it in this PR?

bin/specextract
```diff
@@ -79,381 +81,449 @@ v4 = make_verbose_level(toolname, 4)
...
-            mkacisrmf.wmap = f"{wmap_sky2tdet}[wmap]"
-        else:
-            mkacisrmf.wmap = f"{specfile}[WMAP]" # dmextract WMAP is faster than sky2tdet WMAP
-
-        mkacisrmf()
-        
-        return rmffile
-        
-    else:
+    except (OSError,IOError):
```
DougBurke, line 373: `IOError` is the same as `OSError` in Python 3, so this can just be

```suggestion
    except OSError:
```

nplee:
> @nplee -the first commit is (I'm guessing) the same as #496 - do we need it in this PR?

No.  The two pulls must've been cross-pollinated their commits when I rebased my active branches after the latest contrib update.

bin/specextract
```diff
@@ -79,381 +81,449 @@ v4 = make_verbose_level(toolname, 4)
...
+        arfcorr.clobber = "yes"
+
+        arfcorr()
+
+        # add arfcorr history to file, since it doesn't automatically
+        param_arfcorr = {}
+        for acpar in arfcorr._parnames:
+            param_arfcorr[acpar] = arfcorr._get_param_value(acpar)
+
+        add_tool_history(carffile,"arfcorr",param_arfcorr)
```
DougBurke, line 266: This really should not be needed, but I see it still is...

bin/specextract
```diff
@@ -79,381 +81,449 @@ v4 = make_verbose_level(toolname, 4)
...
-                pbk_kw_list = list(pbk_kw)
+            infile = infile.replace(evt_filename, evt_filename_gz)
 
-                verify_keys = headerkeys.keys() & pbk_kw_list # find elements in common
+            v1("NOTE: {0} does not exist, but {0}.gz does; using the latter as input to arfcorr for the ARF correction.\n".format(evt_filename))
 
-                for s in verify_keys:
-                    pbk_kw_list.remove(s)
+    # arfcorr required input image:
+    reg_image  = f"{infile}[bin sky={binarfcorr}]"
```
DougBurke, line 242: A question about this (and this is the same as what was done before as far as I can tell): would it make sense to add an energy filter here so we don't include particle events. Or does it not really matter?

And also is this only done for ACIS data or also HRC (if the latter then it would just complicate how you would filter the event file)?

bin/specextract
```diff
@@ -79,381 +81,449 @@ v4 = make_verbose_level(toolname, 4)
...
-##########################################################################
-# Usage:
-#  extract_spectra( full_outroot, infile, ptype, ewmap,
-#                   binwmap, instrument, clobber, verbose)
-#
-# Aim:
-#   Create spectrum from input file.
-#
-##########################################################################
+    # NB: Adjust call_calquiz() function eventually
```
DougBurke, line 279: Is this comment still meaningful?

bin/specextract
nplee: good to know.  I've revised this on my local version.

bin/specextract
```diff
@@ -79,381 +81,449 @@ v4 = make_verbose_level(toolname, 4)
...
-        dmextract.wmap = f"[energy={ewmap}][bin {binwmap}]"
-    else:
-        dmextract.wmap = ""
+    caldb_p2_resp_file = calquiz_results[1]
 
-    dmextract()
+    v3(f"Found the following P2_RESP file(s): {caldb_p2_resp_file}")
 
-    return specfile
+    if not numCalFiles or numCalFiles in [0.0, 0, 0.]:
```
DougBurke, line 297: I haven't got to `call_calquiz` yet, but in general I would expect that the return value should be an integer or, even better, you just return a list of names so you can use `len` to identify how many files are matched. If you want to return more-structured data than return something like a dictionary or a namedtuple.

bin/specextract
```diff
@@ -79,381 +81,449 @@ v4 = make_verbose_level(toolname, 4)
...
+        with new_pfiles_environment(ardlib=True,copyuser=False):
+            if rmftool == "mkrmf":
+                mkrmf.infile = feffile
+                mkrmf.outfile = rmffile
+                mkrmf.logfile = ""
+                mkrmf.weights = ""
+                mkrmf.axis1 = f"energy={ebin}"
+                mkrmf.axis2 = rmfbin
+                mkrmf.clobber = clobber
+                mkrmf.verbose = str(verbose)
```
DougBurke, line 343: we shouldn't need str(verbose), it should accept an integer (and I'd claim it's a bug if it doesn't),

bin/specextract
```diff
@@ -79,381 +81,449 @@ v4 = make_verbose_level(toolname, 4)
...
-#              rmfbin, clobber, verbose, specfile, weightfile, ccd_id, chipx, chipy)
-#
-#
-# Aim:
-#   Run either mkrmf or mkacisrmf depending on input conditions.
-#   For mkrmf:  input FEF file and no weight file and return RMF name.
-#   For mkacisrmf:  input CALDB and WMAP and return RMF name.
-#
-##########################################################################
+def build_rmf_ext(rmftool, ptype, full_outroot, ebin, rmfbin, clobber, verbose,
```
DougBurke, line 378: So, without manually going line by line, it looks like this and `build_rmf_ps` could just call a `builf_rmf_basic` routine where you send in the arguments. This may be a bit awkward (passing through too many arguments) so maybe it wouldn't be helpful. Maybe there should be two low-level routines (one for mkrmf and one for mkacisrmf) that are called by `build_rmf_ps` and `build_rmf_ext`.

bin/specextract
```diff
@@ -79,381 +81,449 @@ v4 = make_verbose_level(toolname, 4)
...
-            kval = None
-			
-        keys.append(kval)
-	
-    return keys
+            mkarf.engrid = f"grid({rmffile}[SPECRESP MATRIX][cols ENERG_LO,ENERG_HI])"
+            mkarf()
 
+        except OSError:        
+            mkarf.engrid = f"grid({rmffile}[MATRIX][cols ENERG_LO,ENERG_HI])"
```
DougBurke, line 514: Can we not identify from the `rmffile` argument whether to use [MATRIX] or [SPECRESP MATRIX] rather than have this try/except pattern? I'm sure we have some code that list the block names in a file.

bin/specextract
```diff
@@ -485,244 +556,95 @@ def set_badpix(evtfile, bpixfile, instrument, verbose):
...
-#
-# Aim:
-#  Read the OBS_ID value from the header of each file in the input stack
-#  in order to create a dictionary matching each ObsID to one or a list of
-#  files. Then, this dictionary can be used by other functions in the script
-#  to assign the appropriate file(s) (e.g., asol) to each source observation,
-#  by obsid.
-#
-##########################################################################
+    ds = cr.get_dataset()
```
DougBurke, line 633: rather than g ofrom crate to dataset you should be able to do sokmething like (untested)

```
ds = pcr.CrateDataset(pi, mode='r')   # I'm guessing about mode=r, it may need to be rw
ds.delete_crate('WMAP')
ds.write(tmp_spec_name, clobber=True)
```

bin/specextract
```diff
@@ -828,860 +753,1086 @@ def clip_wmap(wmapfile,threshold,tmpdir):
...
-# Usage:
-#  check_event_stats(file)
-#
-#
-# Aim:
-#     Use dmlist to determine the event counts
-#     and appropriately error out or proceed with the script
-#
-################################################################
+    return dmhedit()
```
DougBurke, line 811: With the code as written you've gone to the effort to setting up a temporary pfiles environment but then it gets replaced before you run the dmhedit call. So either

```
        # this is within the new_pfiles_environment context manager
        result = dmhedit()

    return result
```

or just don't bother with `new_pfiles_environment` here (it's not obvious to me why you would need it).

Actually, is there any place that you use the return value from `edit_headers`?

bin/specextract
```diff
@@ -828,860 +753,1086 @@ def clip_wmap(wmapfile,threshold,tmpdir):
...
 
-    dmlist.punlearn()
 
-    if ewmap_range_check is None:
-        dmlist.infile = file
-    else:
-        dmlist.infile = f"{file}[energy={ewmap_range_check}]"
-        
-    dmlist.opt = "counts"
+def call_calquiz(infile, product, query, file, verbose):
```
DougBurke, line 815: Ahh, the mysterious `call_calquiz`. Is there any reason why we can't just use the caldb4 module? There could be, as I can't remember how it handles being used in multiple places.

bin/specextract
```diff
@@ -828,860 +753,1086 @@ def clip_wmap(wmapfile,threshold,tmpdir):
...
+                try:
+                    ancrfile = create_arf_ps(full_outroot,filename,
+                                             asphist,ebin,clobber,verbose,
+                                             da,msk,
+                                             ccd_id,skyx,skyy,infostr)
+
+                    add_tool_history(ancrfile,toolname,pars,toolversion=__revision__)
+                    
+                except OSError:
+                    raise IOError(f"Failed to create ARF for {fullfile}")
```
DougBurke, line 1028: For these errors (there's a number of them here) we should use either

```
except OSError as exc:
    raise OSError(...blah....) from exc
```

or

```
except OSError:
    raise OSError(... blah ...) from None
```

depending on whether we care about the original error or not (it can be useful when using --debug in our scripts to either b able to see the underlying error or not to have to read the error). I'd suggest going for the first version [i.e. in this case you could get an error either from `create_arf_ps` or `add_tool_history` so you may want to be able to find the original error] and then setting them to `raise .. from None` once you've decided that we don't care about the original error.

bin/specextract
```diff
@@ -828,860 +753,1086 @@ def clip_wmap(wmapfile,threshold,tmpdir):
...
+                    respfile = build_rmf_ps(rmftool,filename,fullfile,ptype,full_outroot,
+                                            ebin,rmfbin,clobber,verbose,phafile,feffile,
+                                            ccd_id,chipx,chipy,infostr)
 
-        if colname == "x":
-            return src_x
+                try:
+                    add_tool_history(respfile,toolname,pars,toolversion=__revision__)
+                except OSError:
+                    v3("skip writing file history to RMF")
```
DougBurke, line 1126: Is it worth this `try/except` block around `add_tool_history` here? I can't think of a case where it should fail but it's okay to carry on, but maybe there's a helpdesk ticket I'm forgetting.

bin/specextract
```diff
@@ -828,860 +753,1086 @@ def clip_wmap(wmapfile,threshold,tmpdir):
...
 
-        elif colname == "y":
-            return src_y
+        except (OSError,IOError):
+            raise IOError(f"Failed to create RMF for {fullfile}")
 
-        elif colname == "chipx":
-            return int(float(src_x))
+    if srcbkg == "bkg" and not dobkgresp:
+        pass
```
DougBurke, line 1132: Another 'did you really mean pass` case.

DougBurke:
@nplee - sorry these comments are 

a) going to come in dribs and drabs
b) but a bit curt

bin/specextract
```diff
@@ -828,860 +753,1086 @@ def clip_wmap(wmapfile,threshold,tmpdir):
...
-    dmgroup.punlearn()
+    try:
+        outreg.close()
+    except NameError:
+        pass
+                
+    try:
+        return pardict
+    except NameError:
+        pass
```
DougBurke, line 1327: This looks really confusing (especially as it's the last line in the function). Why is it okay to ignore a `NameError` and what's meant to happen when it happens?

bin/specextract
```diff
@@ -1823,1511 +1982,123 @@ def specextract(args):
...
-                cs_bkg_spectra = list(range(src_count))
-                cs_bkg_arfs = list(range(src_count))
-                cs_bkg_rmfs = list(range(src_count))
 
-                bphafiles = list(range(src_count))
-                barffiles = list(range(src_count))
-                brmffiles = list(range(src_count))
+            
+if __name__  == "__main__":
+    try:
```
DougBurke, line 2094: If you add

```
import ciao_contrib.logger_wrapper as lw

...

@lw.handle_ciao_errors(toolname, __revision__)
def run_specextract(..):
...
```

then you don't need this `try/except` call.

DougBurke: And I see `run_specextract` has this, so you don't need it.

kglotfelty:
More a question ... since this is now parallelized, shouldn't we add the stock

    parallel,b,h,yes,,,"Run processes in parallel?"
    nproc,i,h,INDEF,,,"Number of processors to use"

parameters to allow the user to control how many cores it uses?

bin/specextract
nplee: revised!

bin/specextract
nplee: not directly from what I can tell, but using some variation of:


```
    blnames = [bl[0].lower() for bl in ciao_contrib.cxcdm_wrapper.get_info_from_file(rmffile)]

    if "specresp matrix" in blnames:
        blname = "SPECRESP MATRIX"
    else:
        blname = "MATRIX"

```

will do the trick to avoid try/except.

bin/specextract
nplee: This works and I've updated things, albeit this might wind up being unused code that can be dropped, now that I've done some further time testing with WMAP clipping that this functionality may go along with.

bin/specextract
nplee: This is also used with PSF-corrected HRC ARFs.  As far as I can tell, arfcorr in ARF-correction mode uses the input image is used to determine the size and scale of the PSF, and the response's energy grid is used for each step of PSF-corrected response.  I'm doubtful that the high-energy particle events would really impact the estimated PSF size.

bin/specextract
nplee: I've modified the function, since the returned value is never used, and sensibly you're correct there's no point to using a ```new_pfiles_environment``` here, since I can't see where there's a possibility of a race condition in its usage.

bin/specextract
nplee: 👍

DougBurke:
Replaced by #535

PR #494, opened 2021-09-16
Chandra repro bowtie not applicable

kglotfelty:
Changes requested by Dale to allow for hrc-s/bowtie filter to be optional.  Pending CALDB update will make it not-applicable for recent (circa 2021) data due to gain/high-voltage something or the other.

This change allows the CALDB query to return 0 results and then just skip the bowtie filtering rather than erroring out (and tripping up of the bad call to error_out, which is also now corrected in several places).

### DIFF
--- a/bin/chandra_repro
+++ b/bin/chandra_repro
@@ -164,9 +164,9 @@ def filter_aspsolobi_from_aspsol(asol_list,mydir,params):
         
     if "ASPSOLOBI" in content_values:
         if len( content_values["ASPSOLOBI"] ) > 1:
-            error_out("Too many ASPSOLOBI aspect file found.  If this "+
+            error_out(params, "Too many ASPSOLOBI aspect file found.  If this "+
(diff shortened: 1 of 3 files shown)


bin/chandra_repro
```diff
@@ -2096,10 +2096,11 @@ def lookup_hrc_pi_filter(evtfile):
     calfiles = cdb.search
     
     if not calfiles or len(calfiles) == 0:
-        error_out("No TGPIMASK2 CALDB files found for this dataset")
+        v1("No TGPIMASK2 CALDB files found for this dataset")
```
DougBurke, line 2099:
```suggestion
        v1("No TGPIMASK2 CALDB file found for this dataset")
```

Technically it's a single file, isn't it?

DougBurke (approved):
LGTM - do you want to make the small change I suggested or not bother?

bin/chandra_repro
kglotfelty: updated.

PR #496, opened 2021-09-21
blanksky_sample: constrain time bounds of blanksky events

nplee:
incorporate deadtime correction to the constrain the time bounds for the randomly assigned values to events and reflect effective ONTIME/LIVETIME/EXPOSURE of reference observation and not just rely on TSTART/TSTOP.

### DIFF
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -18,6 +18,11 @@ New scripts
 
 Updated scripts
 
+  blanksky_sample
+
+    The dead-time correction is now used to better sample the time
+    values created for the events in the blank-sky file.
+
   chandra_repro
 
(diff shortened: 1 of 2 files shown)


DougBurke:
Does this address a helpdesk ticket (ie I seem to remember something about it way back, but maybe mis-rembering)?

nplee:
It doesn't, but it's helpdesk-adjacent.  Found this bug while investigating the ticket to find an alternative approach.  The underlying issue has to do with blanksky_image and I'm not convinced what we're currently doing is incorrect.  I think we can improve things by encoding reference observation info into the blanksky file in a separate block for the script to use, but that's something that probably needs further thinking and discussion.

DougBurke:
Note that the actual change came in in #535 - this is now just a document update.

PR #520, opened 2021-11-01
Increase number of digits shown in scrflux in screen output

hamogu:
While the correct conf setting is included in the output, it is tempting
for the user to just look at the screen output. However, the screen
output is formatted to show just two digits, which is too few for common
values as as 99.7 %. This PR simply increases the number of digits where
needed without messing up the format where two digits are sufficient:
```
>>>for conf in [0.1, 0.9, 0.99, 0.9999]:
>>>    print("{:40s}".format( "Value     {:5g}% Conf Interval".format( conf *100)))

Value        10% Conf Interval
Value        90% Conf Interval
Value        99% Conf Interval
Value     99.99% Conf Interval
```

Note that my editor is set up to automatically strip trailing white space in
Python files. The real change is in line 2433. I can back out the
whitespace changes if needed.

### DIFF
--- a/bin/srcflux
+++ b/bin/srcflux
@@ -52,8 +52,8 @@ class delme( str ):
     root name.  Under normal conditions they are removed automatically
     however, if there is an error then since the threads
     are independent we can't be sure which files were created and
-    what needs to be cleaned up.  
-    
+    what needs to be cleaned up.
+
     Using this delme() class, we id the files that should always be
     removed -- either will be via the normal code paths, or if
     not when the string gets deleted/goes out of scope, the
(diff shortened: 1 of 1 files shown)


DougBurke:
For reference, I think this shows the new and old behavior:

```
>>> for conf in [0.1, 0.9, 0.99, 0.9999]:
...     print("{:40s}".format( "Value     {:5g}% Conf Interval".format( conf *100)))
...
Value        10% Conf Interval
Value        90% Conf Interval
Value        99% Conf Interval
Value     99.99% Conf Interval
>>> for conf in [0.1, 0.9, 0.99, 0.9999]:
...     print("{:40s}".format( "Value     {:3d}% Conf Interval".format( int(conf *100))))
...
Value      10% Conf Interval
Value      90% Conf Interval
Value      99% Conf Interval
Value      99% Conf Interval
```

PR #523, opened 2021-11-02
change how the tarfile is created

DougBurke:
The mk_script_tarfile 

a) now accepts a single argument - `4.14.0` or `4.14.dev` rather than separating out the two labels
b) internally we now create the tarfile manually rather than use setuptools

The latter is the bigger change, and is because of changes to the python packaging ecosystem - e.g. setuptools will remove the install command and there have been a lot changes related to PEP 517. For now I don't think we need the extra logic of setuptools as

1. we don't need to re-create the hashbang lines of the scripts
2. changes to how the tooling worked meant that it became unclear what the best approach was - we can look for CIAO 4.15 along with CXCCM in a better way to do this.

Creating the tarfile manually is - if I remember - how we used to do it. It's not ideal (the existing code is a bit hard-coded so will need to be updated if we add directories for instance), but will hopefully do for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,31 +43,34 @@ a lot of data, and may rely on unreleased versions of CIAO).
 
 ## How to create a tarball for the CIAO CM team
 
+This has **changed** in CIAO 4.14, bith the internals of the script
+and the arguments.
+
 A) a development release
 
-    % ./mk_script_tarfile 4.12 dev
+    % ./mk_script_tarfile 4.14.dev
 
 will create the file
 
-    ciao-4.12-contrib-DEV.tar.gz
+    ciao-4.14-contrib-DEV.tar.gz
 
 with a version file that looks like
 
-    % cat ciao-4.12/contrib/VERSION.CIAO_scripts 
-    scripts 4.12.DEV Friday, December 12, 2014
+    % cat ciao-4.14/contrib/VERSION.CIAO_scripts
+    scripts 4.14.DEV Friday, December 10, 2021
 
 B) a release candidate
 
-    % ./mk_script_tarfile 4.12 2
+    % ./mk_script_tarfile 4.14.0
 
 will create the file
 
-    ciao-4.12-contrib-2.tar.gz
+    ciao-4.14-contrib-0.tar.gz
 
 with a version file that looks like
 
-    % cat ciao-4.12/contrib/VERSION.CIAO_scripts 
-    scripts 4.12.2 Friday, December 12, 2014
+    % cat ciao-4.14/contrib/VERSION.CIAO_scripts
+    scripts 4.14.0 Friday, December 10, 2021
 
 At present there is no system to automatically look up the version
 number.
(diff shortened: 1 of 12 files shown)


mk_script_tarfile
```diff
@@ -1,174 +1,257 @@
 #!/usr/bin/env python
 
-"""
-Usage:
-  ./mk_script_tarfile <number or dev>
+"""Usage:
+  ./mk_script_tarfile <versino string>
```
kglotfelty, line 4: typo

mk_script_tarfile
```diff
@@ -1,174 +1,257 @@
...
-
-For CIAO 4.9 one version is created, which has both Python 2.7 and
-3.5 directory names in it. This may turn out to be a bad idea,
-in which case two versions will be created. This is still the case
-for CIAO 4.10.
-
-Whilst we could automate the choice of number, it seems safer
-to make the user specify it manually.
+the dotted vresino string, which shuold be 4.14.0 or 4.14.dev
+(adjust versino appropriately).
```
kglotfelty, line 10: More creative ways to spell version :)

mk_script_tarfile
```diff
@@ -1,174 +1,257 @@
...
-        sys.exit(1)
+def build_dist(version, dirname, outdir, python="python3.8"):
+    """Build contrib scripts package into a separate temp dir"""
 
-    elif len(dnames) != 1:
-        sys.stderr.write("ERROR: found {} ".format(len(dnames)) +
-                         "directories matching ciao-*.*!\n")
-        sys.exit(1)
+    outdir = Path(outdir)
+    outdir /= dirname
```
kglotfelty, line 41: I'm assuming that `/` is a python 3.9-ism?  This doesn't work with ciao-install 4.14 (py3.8)

README.md
```diff
@@ -43,31 +43,34 @@ a lot of data, and may rely on unreleased versions of CIAO).
 
 ## How to create a tarball for the CIAO CM team
 
+This has **changed** in CIAO 4.14, bith the internals of the script
```
kglotfelty, line 46: `with`

DougBurke: For reference, when you click a line you get this bizarre "+ in a box" icon 

sherpa


if you click it (or press control-g I think, it will create something like

    ```suggestion
    This has **changed** in CIAO 4.14, bith the internals of the script
    ```

and if you edit the text with your suggestion it can actually be added automatically by the reader. e.g.

```suggestion
This has **changed** in CIAO 4.14, with the internals of the script
```

mk_script_tarfile
DougBurke:
```suggestion
  ./mk_script_tarfile <version string>
```

mk_script_tarfile
```diff
@@ -1,174 +1,257 @@
...
-is picked up automatically from the ciao-x.y/ directory).
-
-For CIAO 4.9 one version is created, which has both Python 2.7 and
-3.5 directory names in it. This may turn out to be a bad idea,
-in which case two versions will be created. This is still the case
-for CIAO 4.10.
-
-Whilst we could automate the choice of number, it seems safer
-to make the user specify it manually.
+the dotted vresino string, which shuold be 4.14.0 or 4.14.dev
```
DougBurke, line 9:
```suggestion
the dotted version string, which should be 4.14.0 or 4.14.dev
```

mk_script_tarfile
DougBurke:
```suggestion
(adjust version appropriately).
```

PR #533, opened 2021-11-04
Merging headers

DougBurke:
Fix up some changes between 4.13 and 4.14 which I don't understand but don't have energy to fix.

Note that this needs to be tested with an installed version - both `conda` and `ciao-install` - as the differences could be down to the way I'm testing (using `pip install .`) versus what CXCCM end up doing.

There are a few header-key-rule changes I have just asked @kglotfelty about. Note that these change  "downstream" products - generally the combined exposure maps and related - so it should be okay to remove them.


DougBurke:
I'm thinking we should add

```
OBI_NUM Skip
```

to both merging-rule files

PR #535, opened 2021-11-05
specextract rewrite - 4.14

nplee:
WIP specextract tested against CIAO 4.14 beta 2 addressing @DougBurke's questions and comments in #491.

Still need to address:

> More a question ... since this is now parallelized, shouldn't we add the stock
> 
> ```
> parallel,b,h,yes,,,"Run processes in parallel?"
> nproc,i,h,INDEF,,,"Number of processors to use"
> ```
> 
> parameters to allow the user to control how many cores it uses?


DougBurke:
So does this replace #491 or do we need to take #491 and then this?

nplee:
@DougBurke this replaces #491

DougBurke:
For reference I just rebased this against the master branch.

DougBurke:
:shipit:

PR #538, opened 2021-11-09
fix runtool

kglotfelty:
fix broken (missing) fstring


DougBurke:
Ahhh. That's why I couldn't trigger it - I needed a can-not-be-run-with-@@foo.par tool... Ta mucho. I need to try and add some basic tests of this stuff.

PR #540, opened 2021-11-09
download_obsid_caldb - catch user error

DougBurke:
I used evts.list rather than @evts.list and was worrying why the script errored out with a programmnig error. This catches this
particular error.

### DIFF
--- a/bin/download_obsid_caldb
+++ b/bin/download_obsid_caldb
@@ -18,17 +18,14 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
-from __future__ import print_function
-
 import os
 import sys
-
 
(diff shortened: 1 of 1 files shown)


DougBurke:
@kglotfelty - as this is your script do you have a comment on this (it should only be an issue for use cases when it was going to fail anyway)

kglotfelty:
This looks good.  I'm assuming that the `read_file` read the `evts.lis` file as an ASCII file which is why it didn't throw an Exception there.

DougBurke:
That was my guess too. Given it's a stack file you just get 1 column with n rows, and no metadata...

```
% python -c 'import pycrates; print(pycrates.read_file("evts.list"));'
   Crate Type:        <TABLECrate>
   Crate Name:        evts.list
   Ncols:             1
   Nrows:             6

```

Issue #397, opened 2020-08-11
randomize taskrunner tasks


kglotfelty:
With ordered dictionaries now in 3.7; `taskrunner` tasks are no longer running in random order.

I'm wondering if we can randomize the `taskrunner` queue (either always or make an optional `.run()` argument) ?

Since in eg `srcflux` not all the tasks are added at the same time, randomizing them when the tasks are created wouldn't be as useful as doing it once the queue is populated and the run method callled (methinks)

DougBurke:
I'm not sure how random they ever were, particularly if you have a set of pre-conditions. If I'm not mis-remembering the code, I think there's a section which selects the next task to run, and that could be randomised.I haven't looked at t his for quite a while now so my memory could be faulty.

PR #548, opened 2021-11-16
Randomize tasks

DougBurke:
This is my attempt to address #397 but it's only partial (it doesn't select a random task, it's just that tasks that get added may be randomized, but this does not randomize the whole set).

It is also not turned on by default. To run you would need to change

    foo = TaskRunner()

to

    foo = TaskRunner(randomize_insert=True)

to test this out.

### DIFF
--- a/ciao_contrib/_tools/taskrunner.py
+++ b/ciao_contrib/_tools/taskrunner.py
@@ -1,6 +1,6 @@
 #
-# Copyright (C) 2012, 2015, 2016, 2019, 2020
-#           Smithsonian Astrophysical Observatory
+# Copyright (C) 2012, 2015, 2016, 2019, 2020, 2021
+# Smithsonian Astrophysical Observatory
 #
 #
 #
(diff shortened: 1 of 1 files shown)


DougBurke:
@kglotfelty - if you're bored you could see if this helps `srcflux`, but my bet is it won't.

kglotfelty:
I tried this with `srcflux`, and even after adding `randomize_input=True` , it's still running things in the order they were added.

DougBurke:
Ta for checking. As the commit mentioned this was not a "true" randomisation but a partial one, and only when adding new tasks to the queue, and only "per grouping". The JoinableQueue documentation doesn't mention how items are selected, and if that can be influenced. We could - shooting from the cuff here - subclass this queue so we can add this logic - but that's a much larger change than we want to work on at this time.

DougBurke:
I'm going to leave this open but mark as a draft.

PR #557, opened 2021-11-18
Srcflux merge bugfix, take 2

kglotfelty:
From a helpdesk ticket 23691 ...

When merging results, srcflux uses the `chip_id` value in the .flux file to identify which LIVTIMEn keyword to use.  But for sources right at the edge of the chip, they can be inside the FOV (which includes dither), but have a chip_id value that corresponds to a chip that is not turned on.  This is because the `chip_id` value is compute at the mean (`RA_PNT`, `DEC_PNT`) aspect.

This causes srcflux to fail during the merging with an unhelpful message:

```bash
# srcflux (30 November 2020): ERROR unsupported operand type(s) for *: 'float' and 'NoneType'
```
because in `merge_get_counts` we 

```python
tab.get_key_value("LIVTIME{}".format(cc))
```
without checking that `cc` is an active chip.  In this case, crates returns `None`, leading to the error message above.

In this case we should set `exptime` to `LIVETIME` to be consistent with the per-obi behavior.

### DIFF
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -87,6 +87,14 @@ Updated scripts
     The script now generates FOV files for each reprojected event file
     and a combined file (that includes all the individual shapes).
 
+  srcflux
+      Bug fix when merging results and a source is located at the very
+      edge of a CCD. This bug is trigger when a source is inside the
+      field of view but the source location computed at the mean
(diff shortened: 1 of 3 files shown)


DougBurke:
@kglotfelty - I squashed together your changes and then tweaked the documentation (mainly a completely-unrelated change moving dax to the front of the changelog)

PR #560, opened 2021-11-19
4.14.0 initial release of updated specextract

nplee:
added support for the standard `ciao_contrib` parallelization parameters `parallel` and `nproc`.  ahelp file updated.

### DIFF
--- a/bin/specextract
+++ b/bin/specextract
@@ -34,7 +34,7 @@ Script:
 __version__ = "CIAO 4.14"
 
 toolname = "specextract"
-__revision__ = "11 November 2021"
+__revision__ = "18 November 2021"
 
 
 ##########################################################################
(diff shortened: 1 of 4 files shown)


ciao_contrib/_tools/specextract.py
```diff
@@ -109,6 +110,14 @@ def __init__(self,params):
         self.wmap_clip = params["wmap_clip"]
         self.wmap_threshold = params["wmap_threshold"]
 
+        if params["parallel"].lower() == "no":
```
DougBurke, line 113: I note that this logic is also in `bin/specextract` - I forget with the split up whether this is expected or not?

nplee: it is, there's some lightweight parallelization going on in here.  If we're alright with just forcing the parallelization and not trying to control it, I'd happily remove this logic.

bin/specextract
```diff
@@ -1786,6 +1790,9 @@ def get_par(args):
     #pars["wmap_clip"] = params["wmap_clip"] = paramio.pgetstr(pfile,"wmap_clip")
     #pars["wmap_threshold"] = params["wmap_threshold"] = paramio.pgetstr(pfile,"wmap_threshold")
 
+    pars["parallel"] = params["parallel"] = paramio.pgetstr(pfile, "parallel")
+    pars["nproc"] = params["nproc"] = paramio.pgetstr(pfile, "nproc")
+
```
DougBurke, line 1795: NOT FOR THIS PR, but I would really like to some day replace the strings in the parameter values with Python-appropriate data types - e.g. for `parallel` use bool and `nproc` use int.

ciao_contrib/_tools/specextract.py
DougBurke: Let's leave that for a later cleanup.

PR #568, opened 2021-12-17
copy HRC RMF file to writable version when the 'rmffile' is set

nplee:
This change is intended for the HRC team's usage.  When the `rmffile` parameter is set to something other than 'CALDB' in `create_hrc_resp`, the script attempts to write a file history directly to the read-only file.  The solution is to replicate the behavior when the parameter is set to 'CALDB' and create a writable copy of the specified RMF with the 'outroot' string.

### DIFF
--- a/bin/specextract
+++ b/bin/specextract
@@ -34,7 +34,7 @@ Script:
 __version__ = "CIAO 4.14"
 
 toolname = "specextract"
-__revision__ = "18 November 2021"
+__revision__ = "17 December 2021"
 
 
 ##########################################################################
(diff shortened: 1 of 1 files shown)


bin/specextract
```diff
@@ -1360,6 +1360,8 @@ def resps(args):
                 except OSError as exc:
                     if dobkgresp:
                         raise IOError(f"Failed to create aspect histogram file for {fullfile}") from exc
+                    else:
+                        asphist = None
```
DougBurke, line 1364:
```suggestion

                    asphist = None
```

DougBurke: The idea is that calls like

```
if condition:
    raise RandomError()
else:
    blah()
```

are "more Pythonic" written as

```
if condition:
    raise RandomError()

blah()
```

PR #571, opened 2022-01-12
Correct URL of footprint server

hamogu:
The URL in the code is correct, but the one given inthe comments is not. 

I'm trying to investigate a case where I did not get the correct answer from `find_chandra_obsid`. So, I looked into the docstring to see what exactly the tools does and how to debug. I want to execute the query manually so that I cna see if the error is due to the script (if it is, expect a PR or issue soon) or if it's an error on the footprint server side. Either way, I noticed that the URL given in the dics is wrong and the PR is to correct that.

### DIFF
--- a/ciao_contrib/cda/search.py
+++ b/ciao_contrib/cda/search.py
@@ -24,7 +24,7 @@ Routines that wrap up access to the Simple Image Access (SIA)
 endpoint at the Chandra Footprint Server. As of mid 2018
 the URL is
 
-    https://cxcfps.harvard.edu/cda/footprint/cdaview.html
+    https://cxcfps.cfa.harvard.edu/cda/footprint/cdaview.html
 
 but prior to that it was
 
--- a/share/doc/xml/find_chandra_obsid.xml
+++ b/share/doc/xml/find_chandra_obsid.xml
@@ -33,7 +33,7 @@
       </PARA>
       <PARA>
 	This tool makes use of the
-	<HREF link="https://cxcfps.harvard.edu/cda/footprint/">Chandra Footprint Service</HREF>,
+	<HREF link="https://cxcfps.cfa.harvard.edu/cda/footprint/">Chandra Footprint Service</HREF>,
 	which can also be searched using a web browser.
       </PARA>
       <PARA title="Single argument">


DougBurke:
Defining the "correct" answer from `find_chandra_obsid` - which is really just a thin wrapper around the footprint server, can be surprising!

Some issues include

- whether the location overlaps the footprint that is recorded in the database
- the footprint service may be missing some obsids

which are "not our problem" issues (in that it is either a documentation issue or could be passed to the archive team).

Or maybe there is a problem with the code in `find_chandra_obsid`.

If you run the script with `verbose=5` which should dump the footprint query - e.g. 

```
% find_chandra_obsid 'hcg 68' verbose=5
Running: find_chandra_obsid
  version: 02 November 2020
Querying name resolvers for name=hcg 68
CADC name query: http://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/cadc-target-resolver/find?format=ascii&service=all&cached=true&target=hcg+68
Response from query:
target=hcg 68
service=ned(ned.ipac.caltech.edu)
coordsys=ICRS
ra=208.420408
dec=40.31849584
time(ms)=1574

Position = (208.420408, 40.31849584, 'ICRS')
Found ra=208.420408 dec=40.31849584
    = 13h 53m 40.89792000000301s 40d 19' 6.585023999989517"
Querying Chandra archive for ra=208.420408 dec=40.31849584 radius=0.016666666666666666 inst=all grat=all
Retrieving URL: https://cxcfps.cfa.harvard.edu/cgi-bin/cda/footprint/get_vo_table.pl?strict=1&POS=208.420408,40.31849584&SIZE=0.016666666666666666 timeout=None
Parsed FIELD -> {'name': 'ObsId', 'format': ('int', None), 'description': 'The CXC ObsID for this dataset.', 'ucd': 'ID_MAIN'}
Parsed FIELD -> {'name': 'target_name', 'format': ('char', '*'), 'description': 'The name of the target as given by the observer'}
...
```

This lists the footprint query used

hamogu:
Just a note: The problem I was originally investigating is not actually a bug, it's that my target falls just into the chip gap for one of the observations, so the answer you get depends sensitively on the seach radius which has different defaults in `find_chandra_obsid` vs. the web interface of the footprint server. This PR still stands though since the way I found that required me to find where this script got the data from - and thus the URL given should be the correct one.

PR #581, opened 2022-02-02
Use `plot.backend` instead of `backend` to keep up with recent sherpa changes

### LINKED ISSUE #579: correct_periscope_drift sherpa plotting changes
In this week's CIAOX build (2022-02-01), we are getting an error in `correct_periscope_drift`

`# correct_periscope_drift (6 Feb 2020): ERROR module 'sherpa.plot' has no attribute 'exceptions'`

Sherpa introduced some changes to plotting in this week's build, but I'm not sure which would have introduced this change.

We need to update the script for whatever plotting changes sherpa makes (both short and long term) . @DougBurke or @hamogu do you have any suggestions for a short term fix?

hamogu:
Do not import sherpa plotting backend directly

With https://github.com/sherpa/sherpa/pull/1191 merged, sherpa now supports switching the plotting backend. That means that we shoud not import be backend directly, but use `plot.backend` instead. This change is backwards compatible.

fixes #579

Note: We could make sherpa support the current use, which would always give you the backend that was loaded when sherpa launched. That's probably OK for these scripts since they are typically run as command-line scripts and not in a longer Python session, but I really, really do not want sherpa to have such an easy way to keep references to non-active plotting backends around because things will break in non-obvious ways. That's why I removed e.g. `plot.begin` which was an internal copy to `plot.backend.begin` and would do bad stuff (technical term) if `plot.backend` is assigned something else (a module, an object).

fixes #579

### DIFF
--- a/sherpa_contrib/marx.py
+++ b/sherpa_contrib/marx.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2018, 2019
+#  Copyright (C) 2018, 2019, 2022
 #           Massachusetts Institute of Technology
 #
 #
@@ -37,7 +37,8 @@ Examples
 
 """
 
-from sherpa.plot import HistogramPlot, backend
+from sherpa.plot import HistogramPlot
+from sherpa import plot
 
 from crates_contrib.utils import write_columns
 
@@ -203,9 +204,9 @@ def plot_marx_spectrum(id=None, elow=None, ehigh=None, ewidth=None,
     hplot.title = 'MARX Spectrum: {}'.format(vals['model'])
 
     # LaTeX support depends on the backend
-    if backend.name == 'pylab':
+    if plot.backend.name == 'pylab':
         hplot.ylabel = 'Flux (photon cm$^{-2}$ s$^{-1}$ keV$^{-1}$)'
-    elif backend.name == 'chips':
+    elif plot.backend.name == 'chips':
         hplot.ylabel = 'Flux (photon cm^{-2} s^{-1} keV^{-1})'
     else:
         hplot.ylabel = 'Flux (photon cm^-2 s^-1)'
--- a/sherpa_contrib/marx.py
+++ b/sherpa_contrib/marx.py
@@ -206,8 +206,6 @@ def plot_marx_spectrum(id=None, elow=None, ehigh=None, ewidth=None,
     # LaTeX support depends on the backend
     if plot.backend.name == 'pylab':
         hplot.ylabel = 'Flux (photon cm$^{-2}$ s$^{-1}$ keV$^{-1}$)'
-    elif plot.backend.name == 'chips':
-        hplot.ylabel = 'Flux (photon cm^{-2} s^{-1} keV^{-1})'
     else:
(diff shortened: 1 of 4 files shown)


DougBurke:
There's also changes needed in the radial profile code as I just got the following from a call to `prof_fit()` with CIAO 4.15 beta 1, which includes the new Sherpa code:

```
AttributeError                            Traceback (most recent call last)
File ~/miniconda3/envs/ciao415-beta1/lib/python3.10/site-packages/sherpa_contrib/profiles/__init__.py:1068, in prof_fit(id, model, rstep, rmin, rmax, rlo, rhi, xpos, ypos, ellip, theta, group_counts, group_snr, label, **kwargs)
   1067 try:
-> 1068     plot.begin()
   1069     _prof_fit_plot.plot(**kwargs)

AttributeError: module 'sherpa.plot' has no attribute 'begin'

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
Input In [43], in <cell line: 1>()
----> 1 prof_fit()

File ~/miniconda3/envs/ciao415-beta1/lib/python3.10/site-packages/sherpa_contrib/profiles/__init__.py:1071, in prof_fit(id, model, rstep, rmin, rmax, rlo, rhi, xpos, ypos, ellip, theta, group_counts, group_snr, label, **kwargs)
   1069     _prof_fit_plot.plot(**kwargs)
   1070 except Exception:
-> 1071     plot.exceptions()
   1072     raise
   1073 else:

AttributeError: module 'sherpa.plot' has no attribute 'exceptions'
```

DougBurke:
As an aside, this is going to be a bit annoying as we normally don't have to branch off for the next CIAO release anywhere near this early. It's likely that we'll just keep this as a separate PR that just isn't merged (at least for now), rather than creatig a new branch and merging in all the changes between the two. Let's see.

DougBurke:
Alternatively we can try and write this in such a way we can use this with CIAO 4.14 and 4.15, but I haven't thought about whether this is remotely feasible.

hamogu:
> Alternatively we can try and write this in such a way we can use this with CIAO 4.14 and 4.15, but I haven't thought about whether this is remotely feasible.

I'm sure it's possible, but a little clunky and I don't think it's worth the effort. Leaving this open or branching already is easier.

sherpa_contrib/marx.py
```diff
@@ -203,9 +204,9 @@ def plot_marx_spectrum(id=None, elow=None, ehigh=None, ewidth=None,
     hplot.title = 'MARX Spectrum: {}'.format(vals['model'])
 
     # LaTeX support depends on the backend
-    if backend.name == 'pylab':
+    if plot.backend.name == 'pylab':
         hplot.ylabel = 'Flux (photon cm$^{-2}$ s$^{-1}$ keV$^{-1}$)'
-    elif backend.name == 'chips':
+    elif plot.backend.name == 'chips':
```
DougBurke, line 209: I think we can remove this! GH annoyingly won't allow me to suggest the removal as line 209 wasn't changed in this PR...

Issue #96, opened 2018-02-12
Migrate scripts to new region module


kglotfelty:
CIAO 4.10 includes the rewrite of the region module.  The scripts should be updated to take advantage.
From what I can tell, the following are candidates:

- [ ] `splitroi`
- [x] `src_psffrac`
- [x] `srcflux`
- [X ] ~`reg2tgmask`~
- [X ] ~`ciao_contrib.region.check_fov`~

The old interface is still functional in 4.10 but should be considered deprecated (methinks).

kglotfelty:
n/a for `reg2tgmask` which needs ds9 format including special `tag`s.

DougBurke:
So, `ciao_contrib.region.check_fov` doesn't parse the FOV file as a region - it manually decodes the structure. I'm not sure using the new region module would help in this particular case.

PR #588, opened 2022-03-01
srcflux: replace old region module with new CXCRegion object

kglotfelty:
This is part of #96 

Due to the increased precision there are some numerical diffs in the regression data that all look good.

### DIFF
--- a/bin/srcflux
+++ b/bin/srcflux
@@ -1569,6 +1569,7 @@ def run_specextract( myparams, outroot, srcreg, bkgreg, ii, at_energy ):
         specextract2.badpixfile = myparams.bpixfile
         specextract2.dtffile = myparams.dtffile
         specextract2.energy_wmap = "300:7000"   # For point srcs, bkg not major effect
+        specextract2.parallel = False   # We are already running in parallel
 
         if dec < 0:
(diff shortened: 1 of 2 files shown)


bin/srcflux
```diff
@@ -247,21 +247,18 @@ def validate_src_and_bkg_regions( myroot, infile, tmpdir, src, bkg ):
...
+            read_file(reg)
+            reg = f"region({reg})"
         except:
-            reg2 = "region({})".format(reg)
-            try:
-                rr = regParse( reg2 )
-                reg=reg2
-            except:
-                raise ValueError("Could not parse '{}' as either a region string or a region file".format(reg))
+            pass
```
DougBurke, line 261: Looking at this diff it looks like we've lost the ability to check if region is not a valid region string. Now, this is probably checked for in `regino_to_phys`, but if I'm reading the code correctly you'll just get an error frmo `dmmakereg` rather than the previous, probably-easier-to-understand error.

Or is this case now checked earlier?

kglotfelty: Good point.  You would have gotten something from `dmmakereg` but probably better to catch here.

I've updated the `validate` routine to add additional checks in e5fd4e7

PR #589, opened 2022-03-10
simulate_psf number of rays

kglotfelty:
This adds a new `numrays` parameter to simulate_psf.  When `numiter=INDEF`, the user can specify the minimum number of rays (ie events).  If both `numrays` and `numiter` are specified, then the tool will use the number of iterations.


bin/simulate_psf
```diff
@@ -1280,7 +1308,10 @@ def iteration_loop( pars, obi_info, src_info ):
         file name is stored in a psf.lis file that will get cleaned up
         at the end.
     """
-    for nn in range( int( pars["numiter"] )):
+
+    nn = 0
+    numrays = 0
+    while nn < int(pars["numiter"]) and nn < 10000:
```
DougBurke, line 1314: I would suggest making 10000 a variable (perhaps an optional argument to the routine) or  a constant, but IO'm not sure it's worth it!

PR #596, opened 2022-03-23
Chandra repro run sso_freeze for solar system observations

### LINKED ISSUE #110: RFE: chandra_repro run sso_freeze on ssos (maybe)
Based on [an internal discussion](https://icxc.harvard.edu/pipe/ciaodoc/2017/0161.html) we may want to go ahead and run sso-freeze on solar-system objects since there really isn't anything for user to select nor parameters to change.

This just adds coordinates, doesn't take anything away.

kglotfelty:
This closes #110 

There is no specific meta data in the data products that indicates that this an observation is of a solar system object :frowning_face: 

But `chandra_repro` is looping through the files in the primary directory so we can look for `eph1.fits` files.  All observations will have an `orbit*eph1.fits` file (coded to look for possibility for more than 1 since they are not cut on OBS_ID boundaries).  If `chandra_repro` finds a `eph1.fits` file that doesn't start with `orbit` then we assume we are in a solar-system object scenario.

The columns that `sso_freeze` adds to the event file are benign to other down-stream processing; except `tg_resolve_events` which only copies columns in the eventdef.  So there's a little logic needed to run this after `t_r_e` (when gratings are used -- about 10% of the sso observations to-date).

The sso aspect solution is a little problematic since `sso_freeze` sets the block name to `ocsol` which `asphist` doesn't recognize.  So we cannot change the `ASOLFILE` header keyword in the event files to the sso asol file.

Currently Draft since I need to update the ahelp, but that will cause a merge conflict with #595 .  I may just edit the ahelp on that branch to avoid the headache.

### DIFF
--- a/bin/chandra_repro
+++ b/bin/chandra_repro
@@ -2589,7 +2589,12 @@ def process_sso(pars, is_gratings=False):
     
     v5("sso_freeze returned")
     v5(out)
-
+    
+    # It would be nice to set the ASOLFILE keyword to the sso asol
+    # since it has the same format. Unfortunately, the sso asol 
+    # has blockname="ocsol" instead of "ASPSOL" and that causes
+    # down stream tools (asphist) to error out.  Leave this 
+    # here for the day that's fixed.
     # ~ tab = pyc.read_file(out_evt1, mode="rw")
     # ~ tab.get_key("ASOLFILE").value = os.path.basename(out_asol1)
     # ~ tab.write()
@@ -2600,7 +2605,6 @@ def process_sso(pars, is_gratings=False):
     pars[evt] = out_evt1
     pars["cleanup_files"].append(out_evt1)
     
-
 
 #!############
 # main routine
--- a/bin/chandra_repro
+++ b/bin/chandra_repro
@@ -489,6 +489,8 @@ def make_default_pars():
              "msk1_file":       "",
              "mtl1_file":       "",
              "stat1_file":      "",
+             "eph1_file":       [],
+             "sso_file":        None,
              "pbk0_file":       "",
              "bias0_file":      "",
              "asol1_file":      "",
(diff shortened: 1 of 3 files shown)


kglotfelty:
Hum... I clicked "Ready to review" too quick.

Let me rebase and then update ahelp/changes.

bin/chandra_repro
```diff
@@ -2532,8 +2546,65 @@ def remake_fov1_file( pars ):
...
+    v5(out)
+    
+    # It would be nice to set the ASOLFILE keyword to the sso asol
+    # since it has the same format. Unfortunately, the sso asol 
+    # has blockname="ocsol" instead of "ASPSOL" and that causes
+    # down stream tools (asphist) to error out.  Leave this 
+    # here for the day that's fixed.
+    # ~ tab = pyc.read_file(out_evt1, mode="rw")
+    # ~ tab.get_key("ASOLFILE").value = os.path.basename(out_asol1)
+    # ~ tab.write()
```
DougBurke, line 2600: Could we just change the block name of the file (or write out a copy of it with the name ASPSOL)? I think crates should allow is to do this, but maybe there's other things to worry about?

kglotfelty: So there are two things 

I tried to change the `.name`, but doing that causes crates to write out a new extension to the file -- not update the existing block.  So instead of updating the existing file I'd have to write out a temporary file and deal with that.  Which seems like more trouble than it's worth.

The 2nd reason is that w/ this failing in so early w/ `asphist` it's unclear if there might be other cases further down stream where normal/typical things could go astray because of the extra columns. (There shouldn't be but haven't tested through all the tools that use asol, incl. marx/chart).

DougBurke: Awesome...

PR #617, opened 2022-04-15
change SCRIPTVER/SCRIPTMOD keywords to SCRPTVER/SCRPTMOD

nplee:
As noted by @kglotfelty, FITS header keywords should be <= 8 characters, or the DM writes them as a pair of DTYPE/DVAL keywords which is not optimal.


nplee:
covered by PR #616

PR #622, opened 2022-06-15
fix docstrings in psf_contrib

hamogu:
When using this again over a year after I wrote it, I noticed that some of the docstrings were messed up.


DougBurke:
Ta.

PR #629, opened 2022-08-16
dax: fix wcs when image is blocked

### LINKED ISSUE #570: dax apply blocking to saved image
ds9 doesn't update the WCS when saving which is why dax needs to separately filter the save file to apply any cropping:

https://github.com/cxcsds/ciao-contrib/blob/151f7fc17bc4b3a2dd32513ebf930a3facdde461/bin/ds9_imgproc_wrapper#L184

This also applies to **blocking** (not binning nor zooming). If the image has been blocked the wcs isn't updated when the image is saved which causes coords to be off.  

dax can get the blocking factor same as w/ cropping and apply it to the data when creating the temp file so that things should work better.

kglotfelty:
This fixes #570 .

The original plan was to save the wcs from ds9 (needed 8.4)  and then just dmhedit it into the ds9 temp file.  But it turns out that this can wreak havoc since ds9 applies it's own data-model (probably matches AST's data model).

So instead we are taking a more surgical approach and just updating the scale factors.  The DM adds a linear physical WCS to images that don't explicitly have one already so this also works with non-Chandra/CIAO images.  Updating the `SCALE` updates the `PIX` automatically to preserve the offsets.

As is mentioned in the comments, the DM will impose its own limitations on the WCS -- but whatever CIAO tool is going to be run will do the same.  Images w/ "fancy" WCS's (full skew matrix, -SIP coefs, etc) will continue to be borked.

### DIFF
--- a/bin/ds9_imgproc_wrapper
+++ b/bin/ds9_imgproc_wrapper
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #
-#  Copyright (C) 2019-2021  Smithsonian Astrophysical Observatory
+#  Copyright (C) 2019-2022  Smithsonian Astrophysical Observatory
 #
 #  This program is free software; you can redistribute it and/or modify
 #  it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


DougBurke:
Could this go out for CIAO 4.14 ?

DougBurke:
I assume it requires ds9 8.4 (given the comments in #570).

kglotfelty:
My original approach was tied to ds9 8.4, but this approach isn't.  I would advocate for this waiting until 4.15 anyways to go with the grating interactive coords thing (hopefully).

DougBurke:
I've started the merging - there's no reason for me not to take this is there?

kglotfelty:
Good to go.

DougBurke:
Closed via #652

PR #631, opened 2022-08-30
ahelp: categorize generic tools

kglotfelty:
As per discussion with @hamogu  earlier this year this updates the contrib tools ahelp files to specialize the "tools" content.  The ahelps for the compiled tools will be updated shortly for upcoming ciaox/t.

The end-result is that the ahelp index by context page, https://cxc.cfa.harvard.edu/ciao/ahelp/index_context.html , will then provide better grouping of related tools.

I expect that this will need to be rebased by the time we merge it into the 4.15 build.

> Note: Originally I was using `Tools/group` but `/` causes problem with the publishing scripts.  Same thing if the group had spaces.


DougBurke:
Technically this could go out in 4.14 (I believe) but it makes sense to wait for 4.15

PR #635, opened 2022-09-19
Remove __future__ for print and some unused imports

hamogu:
I was in the process of writing a new tool and wanted to use one of the existing ones as template, when I stumbled on those harmless, but no longer needed lines.
So, I figured, I might as well remove them.


DougBurke:
While we could do this now, I'm going to wait for the CIAO 4.15 release as it's not an essential change.

PR #636, opened 2022-10-03
Add an aplimits tool

hamogu:
This is a draft of the tool for ciao_contrib, hopefully to be included in the 4.15 release. The Python code for the calcualtion itself is tested and works. However, this is the first time I'm writing a new ciao_contrib tool from scratch and JCM strongly encouraged me to put code up for discussion before finalizing choices.

So, here are some open issues to be worked on:

- [?] This tool is somewhat of a companion to `aprates`. One important difference is that it does not allow for a certain fraction of the PSF to fall into the background region. I think that's OK, but we need to decide that.
- [?] Similarly,` aprates` converts into energy fluxes, assuming some average energy. We could do that here, but I have not. Should I?
- [x] I envision an output akin to aprates, i.e. output is written into an output param file that can be queried with `pget`. I don't know exactly how to do that - the current code is untested and probably missing a few lines with parameter descriptions for the output file etc.
- [x] Analog to `aprates` there should be a few parameters that govern the numerics (e.g. max number of evaluations for bisect). Those are currently hardcoded, but it's trivial to make them into parameters. 
- [x] Add description of each parameter to XML file
- [x] Add examples to XML file

It might be good to hook this up with src_flux, maybe though #626 or some other mechanism, so I suggest to read the questions above with that in mind.

## Plan
As I said, I'd like to get this into 4.15, so I'm hoping on some feedback on the questions above soon, so I can finish this work.

There are a few aspects of CIAO tools that I'm less familiar with (the logging system, paramio). I can read the code of other examples and try to figure it out, but if you happen to know how to correct my syntax on top of your head, I'm gladly accpeting suggestions.


hamogu:
While this is a "draft", I do need some feedback from @kglotfelty or @DougBourke to finish this.

param/aplimits.par
```diff
@@ -0,0 +1,11 @@
+typeIerror,r,a,.9,0., 1.,"Upper limit on the probability of a type I error"
+typeIIerror,r,a,.5,0., 1.,"Upper limit on the probability of a type II error"
+T_b,r,a,,0,,"Exposure time in source aperture"
```
kglotfelty, line 3: There are two `T_b` and `A_b`, I'm assuming these should be `T_s` and `T_b`

bin/aplimits
```diff
@@ -0,0 +1,381 @@
...
+                         typeIIerror=float(pars["typeIIerror"]),
+                         T_s=float(pars["T_s"]),
+                         A_s=float(pars["A_s"]),
+                         bkg_rate=float(pars["bkg_rate"])
+                         m=int(pars["m"]),
+                         T_b=float(pars["T_s"]),
+                         A_b=float(pars["A_s"]),
+                         max_counts=int(pars["max_counts"])
+
+    ph = pio.paramopen(TOOLNAME + "_out")
```
kglotfelty, line 375: There are 2 options

1. Either store the values back into the tools own parameter file, like `dmcoords` does or
2. Store the values into a new parameter file like `aprates`

For the `aprates` approach, the paramio module isn't helpful -- it can only be used to set/get values it can't create parameter files from scratch.  So you'd probably want to add an `outfile` parameter and then just something like 

```python
with open(outfile, "rw") as fp:
    fp.write(f's_star,r,h,{sstar},,,"Some comment"\n"')
    fp.write(f'upper_limit,r,h,{ul},,,"Upper limit"\n')
```

You might want to consider saving the PDF as a table (and then these scalar values could be written as keywords).  `aprates` doesn't, but it does print the PDF values when the verbose is >=1.  I don't know of a specific use for it, but it's something Aneta and Frank had asked about at different points.

hamogu: I think that `aprates` approach is more appropriate here. In `dmcoords` all coordinate representations are interchangable, i.e. I can put in X,Y to calculate RA/DEC, or I can put in RA/DEC to calculate X,Y. Here, it's more like `aprates` with input and output parameters.

hamogu:
I think this is ready for review. I suggest to pay particular attention to the `ciao_contrib` specific stuff like logging, modes in the parameter file etc. that I usually don't deal with.

Question: Since we don't run tests in this repro, how do I submit regression tests? The look like this:
```
# Test "bbk_rate known"
# Expected numbers are from Kashyap et al. (2010), Fig 6
# need to match with in 5%
bin/aplimits typeIerror=.1 typeIIerror=.5 bkg_rate=3.0 A_s=1 T_s=1 A_b=1 T_b=1 clobber+
pget aplimit_out.par upper_limit
2.7

bin/aplimits typeIerror=.05 typeIIerror=.5 bkg_rate=3.0 A_s=1 T_s=1 A_b=1 T_b=1 clobber+
pget aplimit_out.par upper_limit
3.7

bin/aplimits typeIerror=.01 typeIIerror=.5 bkg_rate=3.0 A_s=1 T_s=1 A_b=1 T_b=1 clobber+
pget aplimit_out.par upper_limit
5.7

bin/aplimits typeIerror=.1 typeIIerror=.95 bkg_rate=3.0 A_s=1 T_s=1 A_b=1 T_b=1 clobber+
pget aplimit_out.par upper_limit
7.5

# Compare perfect bkg_rate and marginalization in the limit of large numbers
bin/aplimits typeIerror=.1 typeIIerror=.5 bkg_rate=3. A_s=1 T_s=1 A_b=1 T_b=1 clobber+
pget aplimit_out.par upper_limit
2.670164108276367
pget min_counts_detect
5
## For m > max_counts we get an identical result
bin/aplimits typeIerror=.1 typeIIerror=.5 bkg_rate=INDEF m=300 A_s=1 T_s=1 A_b=100 T_b=1 clobber+
pget aplimit_out.par upper_limit
2.670164108276367
pget min_counts_detect
5
## m < max_counts we get a similar, but not identical result
bin/aplimits typeIerror=.1 typeIIerror=.5 bkg_rate=INDEF m=300 A_s=1 T_s=1 A_b=100 T_b=1 max_counts=500 clobber+
pget aplimit_out.par upper_limit
2.5924110412597656
pget min_counts_detect
5

# Background area and exposure time don't matter if bkg_rate is already given
bin/aplimits typeIerror=.1 typeIIerror=.5 bkg_rate=3. A_s=1 T_s=1 A_b=1 T_b=1 clobber+
pget aplimit_out.par upper_limit
2.670164108276367
pget min_counts_detect
5

bin/aplimits typeIerror=.1 typeIIerror=.5 bkg_rate=3. A_s=1 T_s=1 A_b=2.3 T_b=.01 clobber+
pget aplimit_out.par upper_limit
2.670164108276367
pget min_counts_detect
5

# What matters, is the expected n_b, i.e. number of bkg counts in source region
bin/aplimits typeIerror=.1 typeIIerror=.5 bkg_rate=1.5 A_s=2. T_s=1. A_b=2.3 T_b=.01 clobber+
pget aplimit_out.par upper_limit
2.670164108276367
pget min_counts_detect
5

# Same min_counts_detect, but upper limit per time is 5 times lower, because the exposure time is longer
bin/aplimits typeIerror=.1 typeIIerror=.5 bkg_rate=.3 A_s=2. T_s=5. A_b=2.3 T_b=.01 clobber+
pget aplimit_out.par upper_limit
0.5340313911437988
pget min_counts_detect
5

# And the same is true for m instead of bkg_rate
bin/aplimits typeIerror=.1 typeIIerror=.5 bkg_rate=INDEF m=300 A_s=2 T_s=5 A_b=100 T_b=10 max_counts=500 clobber+
pget aplimit_out.par upper_limit
0.7797932624816895
pget min_counts_detect
5

# And the same is true for m instead of bkg_rate
bin/aplimits typeIerror=.1 typeIIerror=.5 bkg_rate=INDEF m=300 A_s=2 T_s=5 A_b=1000 T_b=1 max_counts=500 clobber+
pget aplimit_out.par upper_limit
0.7797932624816895
pget min_counts_detect
5
```

hamogu:
Question: Should I also add this to runtool or do we run mk_runtool independenyl before release to catch all changes?

param/aplimits.par
```diff
@@ -0,0 +1,14 @@
+typeIerror,r,a,.1,0., 1.,"Upper limit on the probability of a type I error"
+typeIIerror,r,a,.5,0., 1.,"Upper limit on the probability of a type II error"
+outfile,f,a,"aplimits_out.par",,,"Filename of output file"
```
kglotfelty, line 3: we should leave the default outfile as blank: ""

share/doc/xml/aplimits.xml
```diff
@@ -0,0 +1,347 @@
...
+      </PARA>
+    </DESC>
+
+    <QEXAMPLELIST>
+      <QEXAMPLE>
+        <SYNTAX>
+          <LINE>unix% aplimits typeIerror=.1 typeIIerror=.5 bkg_rate=1.23</LINE>
+          <LINE>unix% pget aplimits_out.par upper_limit</LINE>
+          <LINE>2.44206</LINE>
+          <LINE>3</LINE>
```
kglotfelty, line 132: Just checking -- Is this example right?  It's only `pget`'ing the `upper_limit` value, but you show two values are returned.

kglotfelty:
w.r.t. the `runtool` (and the `Changes.CIAO_scripts` ), we usually just update these once for the big release since there are likely going to be many updates (including from the CIAO code itself).   When we get to the smaller dot updates, with only a few changes then it's usually safe to update runtool & update the changes and any merge_conflicts are usually small.

DougBurke:
For new tools it might make sense to run the "re-create runtool" script
just so it's easier to test. I need to set up a branch or just mark the
code as 4.15-only from here to do this (the latter option is a lot
easier..).

param/aplimits.par
```diff
@@ -0,0 +1,14 @@
+typeIerror,r,a,.1,0., 1.,"Upper limit on the probability of a type I error"
```
kglotfelty, line 1: would it be more user friendly to rename `typeIerror` and `typeIIerror` something more meaningful like `prob_false_detection` and  `prob_missed_detection` ?

hamogu: Good idea. Will change.

DougBurke:
@hamogu - just a heads up that we need this done by mid November-ish (I forget the exact date) for various reasons

hamogu:
I did the para,eter name change. Any other suggestions?

share/doc/xml/aplimits.xml
```diff
@@ -0,0 +1,347 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE cxchelptopics SYSTEM "CXCHelp.dtd">
+<cxchelptopics>
+  <ENTRY context="tools" key="aprates" refkeywords="poisson net count rate flux
```
kglotfelty, line 4: Change to 

context="tools::statistics" key="aplimits"

Issue #625, opened 2022-06-29
convert_xspec_user_model should use same versions as the installed Sherpa and NumPy.


nplee:
From some testing of `convert_xspec_user_model` in 4.14, I think that requiring the models built against the same version of `Sherpa` and `numpy` already installed is the appropriate than what we're doing presently, which pip installs the latest versions of the packages.  Updating:

https://github.com/cxcsds/ciao-contrib/blob/2f54836783295bfec427733aea1383d55efe698c/bin/convert_xspec_user_model#L740-L749

to something like:

```
from numpy import __version__ as numpyver
'''
    save(setupcfg, out)

    # Create the pyproject.toml file
    #
    out = f'''[build-system]
requires = ["setuptools >= 49.1.2",
            "wheel",
            "numpy == {numpyver}",
            "sherpa == {sherpa.__version__}"
           ]
build-backend = "setuptools.build_meta"
'''
```

should do it.

DougBurke:
For `numpy` I should change it to

                "oldest-supported-numpy",

since there's a backwards-compatble issue. Also, this is what's needed to build the models, and not what is installed. The build is done "somewhat hermetically", so the `sherpa` requirement is always going to be awkward (since it's not going to be able to pick up the installed sherpa version, which comes from CIAO, but is going to install it from PyPI. I think). Need to experiment, but the sherpa requirement is probably sensible

DougBurke:
Sherpa is no-longer required to build (as we grab the needed data from the installed version); this probably breaks some subtle requirement of the python packaging system but let's worry about that when we find it.

For `numpy` it now uses the `oldest-supported-numpy` version. This technically **should** work, so I'm going to close this issue until I'm proven wrong.

PR #639, opened 2022-10-07
convert_xspec_user_model updates

DougBurke:
This may address #625 - at least, it removes sherpa from the required list. It also lets me build against [recent versions - e.g. 2.2 - of relxill](http://www.sternwarte.uni-erlangen.de/~dauser/research/relxill/index.html) although some code changes are still needed.


nplee:
I also tried including cpp files in the `find_file_types` approach and it seemed that the model would compile.  I forget whether it would then fail to import into Sherpa or would fail to fit.

PR #640, opened 2022-10-07
support to build the relxill v1 and v2 models into Sherpa/CIAO 4.14

nplee:
my take to PR #639 and issue #625, with description on its usage at:

https://github.com/sherpa/sherpa/issues/1565#issuecomment-1235772431
> In your CIAO environment, please replace `$ASCDS_CONTRIB/bin/convert_xspec_user_model` with: [convert_xspec_user_model.aug31-relxill](https://github.com/sherpa/sherpa/files/9480021/convert_xspec_user_model.aug31-relxill.txt).
> 
> Assuming you're using the Conda-compilers, then set:
> 
> (t)csh:
> 
> ```
> setenv CC ${CONDA_PREFIX}/bin/x86_64-conda-linux-gnu-gcc
> setenv CXX ${CONDA_PREFIX}/bin/x86_64-conda-linux-gnu-g++
> setenv CFLAGS "-Wno-unused-variable -Wno-array-bounds"
> ```
> 
> bash/zsh:
> 
> ```
> export CC="${CONDA_PREFIX}/bin/x86_64-conda-linux-gnu-gcc"
> export CXX="${CONDA_PREFIX}/bin/x86_64-conda-linux-gnu-g++"
> export CFLAGS="-Wno-unused-variable -Wno-array-bounds"
> ```
> 
> If installing with the CIAO 4.14 binaries and default system compilers instead, it may be necessary to set the `LDFLAGS` variable to `-Wl,-rpath,${ASCDS_INSTALL}/ots/lib` and on my clean RHEL8 system I also had to also explicitly point to the system's libfftw, `"l:libfftw.so.3 -Wl,-rpath,${ASCDS_INSTALL}/ots/lib"`.
> 
> For `relxill 2.1`, in the unpacked directory with the source code you will need to change the line:
> 
> `#include "fftw/fftw3.h"`
> 
> to
> 
> `#include "fftw3.h"`
> 
> in the `common.h` and `Relbase.cpp` files.
> 
> The script can then be run from the source directory:
> 
> ```
> unix% convert_xspec_user_model relxill_v2 lmodel_relxill.dat --clobber
> ```
> 
> which should compile the model, and then import into Sherpa:
> 
> ```
> sherpa> import relxill_v2.ui
> ```
> 
> so you'll be able to then set a source model like `set_source(xsumrelline.rll1)`. An important step is to make sure that the models are available to access the data tables the models are dependent upon. Assuming you've unpacked them in `/path/to/directory/with/relxill/tables/files`, then:
> 
> ```
> sherpa> os.environ["RELXILL_TABLE_MODEL"] = "/path/to/directory/with/relxill/tables/files"
> ```
> 
> will set the location from the models to look in. If you would like to avoid declaring the `os.environ["RELXILL_TABLE_PATH"]` each time you import the model package, you can hard code it on line 119 of `common.h`, changing:
> 
> ```
> #define RELXILL_TABLE_PATH "./"
> ```
> 
> to
> 
> ```
> #define RELXILL_TABLE_PATH "/path/to/directory/with/relxill/tables/files"
> ```

I've tested this with Relxill 1.3, 1.4.3, and 2.1.

### DIFF
--- a/bin/convert_xspec_user_model
+++ b/bin/convert_xspec_user_model
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 #
-# Copyright (C) 2012, 2013, 2014, 2015, 2016, 2020, 2021
+# Copyright (C) 2012, 2013, 2014, 2015, 2016, 2020, 2021, 2022
 # Smithsonian Astrophysical Observatory
 #
 #
@@ -67,7 +67,7 @@ for a description of the model.dat format.
 """
 
 toolname = "convert_xspec_user_model"
-toolver  = "02 November 2021"
+toolver  = "31 August 2022"
 
 import sys
 import os
@@ -78,6 +78,8 @@ import glob
 import argparse
 import importlib
 import shutil
+
+from numpy import __version__ as numpyver
 
 import sherpa
 import sherpa.astro.ui as ui
@@ -151,7 +153,7 @@ running
 """
 
 copyright_str = """
-Copyright (C) 2012, 2013, 2014, 2015, 2016, 2020, 2021
+Copyright (C) 2012, 2013, 2014, 2015, 2016, 2020, 2021, 2022
 Smithsonian Astrophysical Observatory
 
 This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


DougBurke:
@nplee - once we have a version in CIAOX, can you check this gain. For me I now only need to do the "fftw/fftw3.h" -> "fftw3.h" change and I can get relxil 2.2 to build. The changes I made were slightly different to yours, and I'm interested in why you created the FunctionMap stuff as I **don't think, but can not guarantee** that it's needed [but I'm linux-only, and maybe there's some reason to bother with the map stuff].

nplee:
The FunctionMap stuff was from observing how the model is packaged up via `initpackage` in XSpec 12.12.1/HEASoft 6.30.1, which I think may be a bit different than older versions.

DougBurke:
Actually, the FunctionMap stuff has been around for quite a while. It's just not something we need to use (I believe) when interfacing with the XSPEC model library (or at least that's been my assumption/hope for the last few years ;-)

DougBurke:
I am assuming that the fixes we made in the 4.15.x releases are "good enough" to let this be closed.

PR #641, opened 2022-10-11
'blanksky' supports usage of ACIS stowed background files

nplee:
the set of stowed background files for CTI-corrected data sets can be used with the `blanksky` script.

### DIFF
--- a/bin/blanksky
+++ b/bin/blanksky
@@ -19,9 +19,9 @@
 #
 
 toolname = "blanksky"
-__revision__  = "12 September 2022"
+__revision__  = "7 October 2022"
 
-import os, sys, paramio, tempfile, shutil, stat, glob
+import os, sys, paramio, tempfile, shutil, stat
 
 import stk
 import pycrates as pcr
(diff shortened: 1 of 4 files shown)


bin/blanksky
```diff
@@ -207,12 +206,119 @@ def check_tool_par(evt,tool,param):
...
+        raise IOError(f"{evtfile} is missing the 'CTI_APP' header keyword!")
+
+    if "N" in "".join(set(ctiapp)).upper():
+        raise IOError(f"The '{toolname}' script only supports CTI-corrected event files for ACIS stowed background file selection.")
+    
+    caldb = os.environ["CALDB"]
+
+    bkg_path = f"{caldb}/data/chandra/acis/bkgrnd"
+
+    epoch_dict = {"Epoch D" : {"tstart" : 92016001.0,
```
DougBurke, line 295: As an aside, I wonder if we can just query the CALDB to find the necessary data (but then it might be hard to do fall back/forward cases).

nplee: One would hope, but I checked.  There's no unique keyword or product query type  that's unique to the stowed background, they're indistinguishable from regular blanksky files, and the caldb4 (BKGRND) query returns only blanksky files.

DougBurke: Joy...

PR #648, opened 2022-10-31
psf_contour (new tool)

kglotfelty:
A new tool to create source region files by simulating PSF and generating regions designed to enclose a fraction of the PSF.  (Filling in one of the gaps w/ acis_extract).

There are several options for choice of region: contour, lasso, ellipse fitted to contour, convex hull around contour, or ellipse determined directly from PSF.

Overlapping regions are dealt w/ similar to AE:  the regions are shrunk (by lowering PSF fraction) until limit reached and then just exclude overlapping regions.


kglotfelty:
TODO: add `fovfile`

PR #649, opened 2022-10-31
Add informative error messages for tools without parameter files

### LINKED ISSUE #643: cannot use acis_clear_status_bits (and some others) from runtool
A co-worker reported the following problem to me:

I like scripting things from Python, e.g. downloading data goes through `ciao_contrib.cda`, while most tool go through `runtool`. Where is `acis_clear_status_bits`?

After some investigation, I think I can say that `acis_clear_status_bits` is nowhere - it can only be called with `subprocess.call`, not with `runtool`. That's not a big problem in principle (since runtool is just a fancy way of building up the parameter file and then calling the CIAO tool), but it's a problem of discoverability: It's hard to find where things are.

I had, naively, expected to find `acis_clear_status_bits` in runtool, since it has the same flavor as many other contributed scripts, but it's not there, because it does not use a parameter file, instead it's directly reading from `sys.argv`: https://github.com/cxcsds/ciao-contrib/blob/151f7fc17bc4b3a2dd32513ebf930a3facdde461/bin/acis_clear_status_bits#L104
Thus, it's not made into a runtool, which is build by parsing the .par files.

I see three possible solutions:

- Do nothing. No par file wrangling needed, let users call `subprocess.call` directly. I think this is not ideal because of discoverability.
- Add a .par file for `acis_clear_status_bits`. That's backwards compatible because it only has one parameter that is passed in the command line. One would still do that even if it has a par file.
- Add a new category to the existing `CIAOToolDirect` and `CIAOTollParFile` classes, called `CIAOToolNoParFile`.

I note that there are some other tools that are not available in runtool, e.g. `check_ciao_version`. 

The following code lists a number of missing tools:
```Python
import os
from glob import glob
from ciao_contrib import runtool as rt

rttools = set(rt.list_tools())
binlist = set([os.path.basename(f) for f in glob(os.environ['ASCDS_CONTRIB' + '/*')])

binlist - rttools
```
lists

{'acis_clear_status_bits',
 'check_ciao_caldb',
 'check_ciao_version',
 'convert_xspec_script',
 'convert_xspec_user_model',
 'download_chandra_obsid',
 'ds9_aper.sh',
 'ds9_functs',
 'ds9_imgfit.sh',
 'ds9_imgproc_wrapper',
 'ds9_lut',
 'ds9_plot_blt',
 'ds9_radial_fit',
 'ds9_snap2.sh',
 'ds9_specfit.sh',
 'ds9_tgcoord.sh',
 'install_marx',
 'splitroi',
 'summarize_status_bits',
 'tg_bkg

Of those, I would argue that acis_clear_status_bits, splitroi, summarize_status_bits, and tg_bkg should be available in runtool in some form, possibly also the check_* and download_chandra_obsid.

hamogu:
Users (and that includes me before I became part of the team that works on `ciao_contrib`)
don't know and should not have ot care how exactly individual CIAO tools are implemented:
As shell scripts, as C code, as Python code, with or without use or parameter files.

This, users may reasonable expect that `from ciao_contrib import *` gives them access to all the CIAO tools that they find listed on the ahelp pages.
That is not the case and there are good reasons for that (see discussion in https://github.com/cxcsds/ciao-contrib/issues/643#issuecomment-1294078188 .
This PR adds dummys with the names of the missing tools that at the very least will issue a message that the tool is missing and provide some re-direction of what do do instead.

There is the potential downside that now we *do* have e.g. `rt.download_chandra_obsid`, but it dies not do anything. However, I believe it's far more likely (and that's borne out by the questions I've seen so far) that users blindly expect to find all CIAO tools in `runtool` than that they carefully scrutinize the list of module level symbols.

PR includes minor typo fixes in documentation that I reviewed in order to write this code.

closes #643

### DIFF
--- a/mk_runtool.footer
+++ b/mk_runtool.footer
@@ -10,4 +10,57 @@ for toolname in list_tools():
     setattr(sys.modules[__name__], toolname, make_tool(toolname))
     __all__.append(toolname)
 
+# Now, some tools added that don't use par files and are thus not
+# handled by this tool.
+# However, that's a fine distinction that users might not appreciate.
+# Thus, we add them here, so that least there is a helpful message
+# if they try to use it.
+# The list of tools is manually curated and believed to be correct
+# as of COA 4.14 (see https://github.com/cxcsds/ciao-contrib/issues/643#issue-1419049281)
+# We are not adding all scripts, just those that a casual user might expect here.
+# In particular, ds9* and install_marx clearly interact with non-Python systems
+# and as such we believe that users would not look for thme here.
+no_par_file_tools = [
+    'acis_clear_status_bits',
+    'check_ciao_caldb',
+    'check_ciao_version',
+    'convert_xspec_script',
+    'convert_xspec_user_model',
+    'splitroi',
+    'summarize_status_bits',
+    'tg_bkg'
+]
+
+def make_no_par_file_message(toolname):
+    '''Make a dummy function that raises an error with a helpful message when called.'''
+    def no_parfile(*args, **kwargs):
+        '''CIAO tools without parameter file.
+
+        This is a dummy function that will raise an error when called and
+        advice to call the CIAO tool using
+        `subprocess.run()` from the Python standard library.
+        '''
+        raise NotImplementedError(f"{toolname} does not use a parameter file and thus is not handled " +
+                                  "by ciao_contrib.runtool. Use `subprocess.run` instead " +
+                                  f"(see https://docs.python.org/3/library/subprocess.html) to call {toolname}.")
+    return no_parfile
+
+for toolname in no_par_file_tools:
+    setattr(sys.modules[__name__], toolname, make_no_par_file_message(toolname))
+    __all__.append(toolname)
+
+__all__.extend(['no_par_file_tools', 'make_no_par_file_message'])
+
+# And this one is special, so list individually.
+def download_chandra_obsid(*args, **kwargs):
+    '''Dummy function that raises an error when called.
+
+    The error message tells the user to use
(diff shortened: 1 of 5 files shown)


mk_runtool.footer
```diff
@@ -10,4 +10,57 @@ for toolname in list_tools():
     setattr(sys.modules[__name__], toolname, make_tool(toolname))
     __all__.append(toolname)
 
+# Now, some tools added that don't use par files and are thus not
```
DougBurke, line 13: We could add a marker to these scripts [*], so this would change the onus from this script to whoever writes the script-with-no-parameter file. But I'm not sure it's worth it (i.e. if we do this then it's probably best to have the logic here).

[*] ie some easily greppable token

hamogu: Yeah, but then we have to make sure that marker is in each of them, which seems like a thing we'll forget about two years form now and we can't automatically enforce without even more complex logic. We also *could* have the logic from [the codeblock I used to identify those scripts](https://github.com/cxcsds/ciao-contrib/issues/643#issue-1419049281) here, but I decided that the risk of including too much of that accidentally is too high, e.g. I don't think `install_marx` makes much sense here.

I could add a check to make sure that anything added in this stage is not in `sys.modules[__name__]` already, in case we ever add a par file to one of the scripts that don't have it right now.

hamogu: My impression is that this is list is pretty stable because (a) we don't add scripts all that often and (b) when we do most of them use parameter files, so I suggest to start with the simple thing (a hardcoded list) for now. If we find ourselves changing that list often, then we can go and do something more complicated later.

PR #654, opened 2022-11-04
update mk_script_tarfile python version

nplee:
changed to use python3.9 as the referenced directories to match the CIAO 4.15 default version.


DougBurke:
@nplee - can  you rebase this just to get the python change and not the `parallel_pool_futures` stuff?

PR #668, opened 2022-11-08
added 'parallel_pool_futures'

nplee:
method that makes use of `concurrent.futures` as another approach to asynchronous parallelization, since `parallel_pool` is susceptible to hanging if a large number of jobs submitted to it. 

Also provide an approach to displaying a progress bar as parallel_pool_futures completes jobs; the functions may also be used independently with iterators. 

Future `specextract` PRs will include this version of `parallel_wrapper.py`.


DougBurke:
This has been merged in #676.

PR #669, opened 2022-11-09
bkg_fixed_counts: remove the unnecessary inner ECF circle from the background region

kglotfelty:
The inner radius of the annulus already accounts from the inner ecf region.   While technically it does not cause any problems, it looks odd to have an excluded region that isn't actually excluding anything.  Or another way to look at it ... it looks like the same area is excluded twice.

### DIFF
--- a/bin/bkg_fixed_counts
+++ b/bin/bkg_fixed_counts
@@ -26,7 +26,7 @@ import ciao_contrib.logger_wrapper as lw
 
 
 toolname = "bkg_fixed_counts"
-__revision__ = "31 October 2022"
+__revision__ = "09 November 2022"
 
 lw.initialize_logger(toolname)
 verb0 = lw.get_logger(toolname).verbose0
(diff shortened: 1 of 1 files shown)


bin/bkg_fixed_counts
```diff
@@ -173,6 +173,29 @@ def find_outer_bkg_radius(coords, bkg_x, bkg_y, numcts, maxradius):
         coords['outer_rad_pix'].append(outer_rad)
 
 
+def _remove_inner_circle(background, inner_circle):
+    """The ecf circle with the radius equal to the annulus'
+    inner radius is always excluded.  We don't want that.
+    """
+    from region import CXCRegion, opAND, opOR, opNOOP
+    idx = background.index(-inner_circle)
```
DougBurke, line 181: Sp, by construction we know that inner_circle is going to be an exclude region in the background, which is why we don't need to check for the `index` method raising a `ValueError`?

PR #675, opened 2022-11-09
WIP:  srcflux: add aplimits tool to output

kglotfelty:
This PR runs @hamogu `aplimits` tool and includes the output into the `srcflux` output `.flux` file.

The output values are currently being written as

- `min_counts_detect` added as `SRC_MIN_COUNTS_DETECT`
- `upper_limit` gets added as `SRC_UPPER_LIMIT` (there is already an `UPPER_LIMIT` column)

Based on discussion and doc update I think the later should be `SRC_RATE_UPPER_LIMIT` with units `cts/s`.

Questions:

- Should we add `prob_false_detect` and `prob_missed_detection` to the `srcflux.par` file and pass those values to `aplimits` or are the defaults sufficient?

TODO:
- [ ] Store the `prob_*` values (whether defaults or passed in), probably as header keywords in `.flux` file.
- [ ] Apply the psf fraction, `alpha`,  correction to the upper limits
- [ ] Doc updates

Anything else?

### DIFF
--- a/bin/srcflux
+++ b/bin/srcflux
@@ -959,6 +959,7 @@ def run_aprates( c_s, a_s, alpha, c_b, a_b, beta, exposure, conf, outfile ):
     that can be combined together later
     """
     aprates = make_tool("aprates")
+    aplimit = make_tool("aplimits")
 
     def write_null( outfile ):
         with open(outfile, "w") as fp:
@@ -966,6 +967,8 @@ def run_aprates( c_s, a_s, alpha, c_b, a_b, beta, exposure, conf, outfile ):
             fp.write("src_rate_err_lo,r,h,INDEF,,,\n")
             fp.write("src_rate_err_up,r,h,INDEF,,,\n")
             fp.write("src_rate_signif,r,h,INDEF,,,\n")
+            fp.write("min_counts_detect,i,h,INDEF,,,\n")
+            fp.write("upper_limit,r,h,INDEF,,,\n")
         return
 
     verb1( "Running aprates for {}".format(outfile) )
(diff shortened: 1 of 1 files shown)


DougBurke:
What is the runtime hit like here? I know that `ap<I forget it's name>` can be quite slow; is `aplimits` faster?

kglotfelty:
TBD.

FWIW: `aprates` is only slow when users use ... let's say "unrealistic" inputs ... which is less likely via dax.

hamogu:
> What is the runtime hit like here? I know that `ap<I forget it's name>` can be quite slow; is `aplimits` faster?

aplimits is fast enough that u never bothered to measure.  << 1s per source on my machine.

DougBurke:
I'm guessing we don't want to take this for this release. I don't think it's that far away from being mergeable but just not enough time...

kglotfelty:
Agree ... out of time to iron out the remaining wrinkles.  In #684 I added an example plugin that uses `aplimits` which I suggest we take for 4.15 and we an revisit this branch later.

bin/srcflux
```diff
@@ -1091,13 +1127,13 @@ def add_aprates_to_output( myparams, at_energy, outfiles):
...
 
-    def pull_par_into_crate( parname, cratename, desc, units="counts/s" ):
+    def pull_par_into_crate( parname, cratename, desc, dtype=float, units="counts/s" ):
         from paramio import pget
         from pycrates import CrateData
         cd = CrateData()
         cd.name = cratename
         vals = [pget(pp,parname) for pp in outfiles ]
-        vals = [ np.nan if x == 'INDEF' else float(x) for x in vals ]
+        vals = [ np.nan if x == 'INDEF' else dtype(x) for x in vals ]
```
DougBurke, line 1136: So, NaN isn't technically valid for an array of ints. Is crates clever enough to handle this case (and is it clever enough to identify that a list of ints means we can use an integer output type, I mean I assume it is or else this change would not have been made)?

kglotfelty: Good catch.

Yes, crates "casts" the list to a numpy array and uses it's datatype, and numpy is smart enough to check all the values and if they're all int's will create an int column.  If there'd happen to be a NaN, then it'd get cast as a float ... so not what we want.

We'd have to set a NULL value and deal w/ it that way.

kglotfelty:
I'm going to leave this as an optional plugin for users to run.

Issue #672, opened 2022-11-09
specextract: can remove arfcorr  history


DougBurke:
arfcorr now (in CIAO 4.15) includes the history block in the output file, so whatever specextract was doing it can now stop

nplee:
I've applied this change and have regtested it in CIAOT.

Issue #671, opened 2022-11-09
specextract talks about using mkacisrmf twice


DougBurke:
This is with CIAOX but not sure it's changed much.

```
% specextract primary/acisf27294N001_evt2.fits.gz"[sky=region(ds9.reg)]" ciaox/src weight- correct+ clob+ 
Running specextract
Version: 10 February 2022

Aspect solution file primary/pcadf27294_000N001_asol1.fits found.

Bad pixel file primary/acisf27294_000N001_bpix1.fits.gz found.

Mask file secondary/acisf27294_000N001_msk1.fits found.

Setting bad pixel file 

Converting source region to physical coordinates 

Extracting src spectra 

Using mkacisrmf...

Creating src ARF 

Using mkacisrmf...

Creating src RMF 

Calculating aperture correction for src ARF 

Grouping src spectrum 

Updating header of ciaox/src.pi with RESPFILE and ANCRFILE keywords.

Updating header of ciaox/src_grp.pi with RESPFILE and ANCRFILE keywords.
```

Note that the screen output says `Using mkacisrmf...` and then `Creating src ARF` and then `Using mkacisrmf...`. Are we doing redundant work or are strange things afoot at the circle K?

nplee:
@DougBurke There's definitely some redundancies here.  I've corrected this for 4.15, and I'm like 99.999% sure (can't check right now, internet's gone down in the area, but can see the Comcast bucket truck down the street doing work) this is also fixed in the 4.14.4 update.

DougBurke:
With a 4.14.4 run I saw

```
Converting source region to physical coordinates 

Extracting src spectra 

Using mkacisrmf...

Creating src RMF 

Creating src ARF 

Calculating aperture correction for src ARF 

```

so as you said it looks to be fixed there

Issue #660, opened 2022-11-07
specextract clobber checking


DougBurke:
I was surprised at this failure:

```
% specextract repro/acisf27282_repro_evt2.fits"[sky=region(ds9.reg)]" xx/zz correctpsf+ weight- 
Running specextract
Version: 27 October 2022

Aspect solution file repro/pcadf27282_000N001_asol1.fits found.

Bad pixel file repro/acisf27282_repro_bpix1.fits found.

Mask file repro/acisf27282_000N001_msk1.fits found.

Setting bad pixel file 

Converting source region to physical coordinates 

Extracting src spectra 

# specextract (27 October 2022): ERROR Failed to extract spectrum for repro/acisf27282_repro_evt2.fits[sky=region(/tmp/tmp2z0pajr8_phys_coords_src1.reg)]
```

I tracked it down to the fact that the output file already exists, so what I needed to add was `clobber+`. We should review the `specextract` code to see on checking the clobber setting before running the tools .

nplee:
I'll note that I see this also in contrib 4.14.3.

Issue #551, opened 2021-11-17
specextract response position 


nplee:
I've been looking into how specextract currently determines the location where responses are generated if a `refcoord` position isn't provided.  An extreme example is looking at a large extended source:

![CasA-resp_pos](https://user-images.githubusercontent.com/23193862/142230011-0385d02d-7dc2-4770-968e-77b3991f935a.png)

The ellipse is the defined extraction region for Cas A and the magenta 'x' represents the maximum location found by `dmstat`, which is where specextract would presently use to generate an unweighted response.  The cyan mark represents the centroid location determined by `dmstat`, but it appears the reason why the the maximum was picked over the centroid is that dmstat could fail to locate a centroid, and in the past this would cause specextract to fail due to the problem corresponding to zero counts.

The yellow mark presents the calculated center from the extraction region's extent using `region.CXCRegion`, and it looks like the centroiding does a pretty decent job.

![moritz-resp_pos](https://user-images.githubusercontent.com/23193862/142242548-b69a2186-5b63-4171-9ecf-71dd0e18a70a.png)

But in the `arfcorr` issue reported last week, where the user is looking at 90% ECF regions, we see that the maximum and centroid location are identical, but the derived center (yellow) is quite different.  Using the centroid position for the ARF, the PSF_FRAC values are roughly 60% while the calculated center is the expected 90%.

So the question is, should the `dmstat` approach be abandoned in favor of the extent-derived location?

DougBurke:
Something of an aside, but for extended sources we don't care about `arfcorr` (at least for very-extended; for slightly-extended there could still be a reason to include the `arfcorr` correction).

kglotfelty:
FYI -- here is a notebook with a realistic example where we do care about the `arfcorr` correction.

https://nbviewer.org/urls/hea-www.harvard.edu/~kjg/Specextract_Position_Experiments.ipynb

nplee:
The method I've come up with, but haven't pushed yet, is to use the centroid, and if it's within n pixels away from the center of the bounding box, it'd use that position for the responses, otherwise it would use the center of the bounding box.

But if using the center of the first region of a composite region is better, I can be convinced to revise the approach.

DougBurke:
@nplee can we close this or do you want to leave it open?

PR #676, opened 2022-11-09
specextract 4.15.0

nplee:
This PR builds on top of PR #668 and the 4.14.4 version of `specextract`.  This is the initial version to make use of `concurrent.futures` via `parallel_pool_futures` in `parallel_wrapper.py` to manage the asynchronous processing pool and decouple ARF and RMF creation when possible, so they may be generated as independent processes.  

Progress bars will be displayed to monitor processing progress when verbose=1, otherwise, screen output will be more or less identical to earlier versions of specextract at higher verbosity.

The `resp_pos` parameter is also introduced to determine the method of where on the detector unweighted responses are generated at (#551); the default is `resp_pos=REGION` which assumes the center defined in the provided extraction region is the appropriate location while `resp_pos=MAX` will replicate pre-4.15 behavior.  This parameter will be superseded if the `refcoord` parameter is defined.

Internal changes include re-working of how histories are written to files and variable/function names and dictionary usage to avoid race conditions and corrupting parameter parameter files while running in parallel.

This PR also addresses #660, #671, #672.


DougBurke:
Unfortunately I'm not a fan of the new screen output. Here's a 4.14 run (note I'm saving to a file, so it's in the "not a TTY" case):

```
% specextract 9408/primary/acisf09408N002_evt2.fits.gz"[sky=region(ds9.reg)]" ospec/fred clob+ > old.out
% cat old.out 
Running specextract
Version: 27 October 2022

Aspect solution file 9408/primary/pcadf313765973N002_asol1.fits found.

Bad pixel file 9408/primary/acisf09408_000N002_bpix1.fits found.

Mask file 9408/repro/acisf09408_000N002_msk1.fits found.

Setting bad pixel file 

Extracting src spectra 

Using mkacisrmf...

Creating src ARF 

Creating src RMF 

Grouping src spectrum 

Updating header of ospec/fred.pi with RESPFILE and ANCRFILE keywords.

Updating header of ospec/fred_grp.pi with RESPFILE and ANCRFILE keywords.

```

If I run this now I get

```
% specextract 9408/primary/acisf09408N002_evt2.fits.gz"[sky=region(ds9.reg)]" bob/fred clob+ >& new.out
% cat new.out
Running specextract
Version: 9 November 2022

```

This is likely a problem for people who want to review what the script did/record things.

nplee:
@DougBurke try running with verbose=2, which should provide the tool-level verbose=1 output you're expecting.

DougBurke:
Hmm, the verbose=2 output is a touch more opaque than the old verbose=1.

So, when verbose=1 and there's a TTY then you get the progressbar, otherwise if not a TTY then essentially no output. If verbose > 1 then there's no progressbar, whether it's a TTY or not. And verbose=0 there's essentially no output (other than the new high-FP temp warning), so no progressbar there. In short, you only get a progressbar when verbose=1 and it's a TTY. Is that correct?

nplee:
> In short, you only get a progressbar when verbose=1 and it's a TTY. Is that correct?
Correct.

I think I have a tweak within the main `specextract` script that should allow for something more on the lines of the old verbose=1 output when there's no TTY.

DougBurke:
For fun I tried the verbose=1 output with wide and narrow terminals, and it seemed to work okay (when very narrow it's not particularly useful, but that's not something I think we need to worry about).

DougBurke:
It strikes me that we could, in an infinite-free-time sort of way, to rework the parallel code to use the approach taken by `merge_obs` and the like: that is, set up  all the code we want and then run it in parallel (eg `ciao_contrib._tools.taskrunner`). This might be slightly more efficient, as we don't do things like "process these two files and wait to complete, then process these four files, then ...", but it would need a lot of work to see if it even made sense, and definitely not a high priority.

bin/specextract
```diff
@@ -1689,7 +1775,7 @@ def groupspec(args):
             raise IOError(f"Failed to group spectrum for {fullfile}")
```
DougBurke, line 1775: I know this line want's changed, but `fullfile` isn't a defined variable in this routine.

DougBurke: It's also an issue for line 1787.

DougBurke: And it would pobably make sense to change these from

```
    except OSError:
        raise IOError(..)
```

to

```
    except OSError as err:
        raise IOError(...) from err
```

as you get a slightly-more-helpful error trace if they do fail.

bin/specextract
```diff
@@ -2079,34 +2273,55 @@ def run_specextract(args):
             raise IOError(f"The absolute path for the {parname}, '{os.path.abspath(params[parname])}', cannot contain any spaces")
 
     specdicts = spec.ParDicts(params).specextract_dict()  
+    
+    specextract_args = [{"key_id" : key, "key_id_total" : len(specdicts),
+                         "ncores" : nproc , **specdicts[key], "pars_specextract" : pars} for key in specdicts.keys()]
```
DougBurke, line 2278:
```suggestion
                         "ncores" : nproc , **arg, "pars_specextract" : pars} for key,arg in specdicts.items()]
```

Not checked, but it should work.

bin/specextract
```diff
@@ -1999,6 +2191,7 @@ def update_param_args_and_dict(dict_update,dict_orig,args_list,cmd_pars_history)
     else:
         return args_list
```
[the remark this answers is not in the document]
DougBurke: `pylint` likes to flag up

```python
    if cond:
        return somthing
    else:
        return something
```

to point out the `else` is not needed (well, as long as you remove the indentation from the second return call).

bin/specextract
```diff
@@ -1603,34 +1648,75 @@ def resps(args):
...
-            fef_file = None
-
-        try:
-            outreg.close()
-        except NameError:
-            outreg = None
-
     if srcbkg == "bkg" and not dobkgresp:
         return None
     else:
```
DougBurke, line 1653: We can remove the `else` line and de-indent the following lines. Also, are we guaranteed that mkresp is one of these three values? If so the last one can remove the if check (maybe changing it to a 

    assert mkresp == "ARF+RMF"

to catch if extra values can be added to `mkresp`. Alternatively the routine could just end in a

    raise ValueError(f"Unknown mkresp={mkresp}")

bin/specextract
```diff
@@ -1389,21 +1409,91 @@ def spectra(args):
...
+                    asphist = None
+
+        kdict["asphist"] = asphist
+
+        if instrument == "ACIS":
+            rmftool = args_dict["rmftool"]
+                        
+            kdict["tdetwmap"], kdict["feffile"] = resp_setup(args_dict,asphist,rmftool)
+
+        return {key_id : kdict}
```
DougBurke, line 1488:
```suggestion
    return {key_id : kdict}
```

Techincally we don't have to do this, but it makes it a bit clearer when the return happens.

bin/specextract
```diff
@@ -1331,7 +1351,7 @@ def spectra(args):
                 readout_streak_spec(specfile=phafile,infile=fullfile,src_x=float(args["skyx"]),src_y=float(args["skyy"]))
 
             try:
-                pardict[key_id].update({"phafile" : phafile})
+                pardict[key_id].update({"phafile" : phafile})                    
             except NameError:
                 pardict = {key_id : {"phafile" : phafile}}
```
DougBurke, line 1357: The following should be changed to

```python
    except OSError as exc:
        raise IOError(f"Failed to extract spectrum for {fullfile}") from exc
```

although I think this should be moved closer to where the error is going to be raised , but that might be too much of a change now.

bin/specextract
```diff
@@ -1085,31 +1118,33 @@ def build_acis_arf(args):
...
+                    else:
+                        v3(f"Calculating aperture correction for {srcbkg} ARF {iteminfostr}")
 
                     try:
                         ancrfile = correct_arf(full_outroot,fullfile,filename,
                                                ancrfile,skyx,skyy,binarfcorr,clobber)
-                    except OSError:
+                    except Exception as exc:
+                        print(exc)
                         raise IOError(f"Failed to PSF correct the ARF: {ancrfile}")
```
DougBurke, line 1144:
```suggestion
                        raise IOError(f"Failed to PSF correct the ARF: {ancrfile}") from exc
```

bin/specextract
```diff
@@ -975,30 +1014,29 @@ def resp_setup(args,asphist):
...
                 sky2tdet()
 
                 # genearte clipped WMAP to speed up response generation
                 if wmap_clip:
-                    clip_wmap(tdetwmap.name,wmap_thresh,tmpdir)
+                    clip_wmap(tdetwmap,wmap_thresh,tmpdir)
 
             except OSError:
-                tdetwmap.close()
                 raise IOError("Error generating WMAP w/sky2tdet!")
```
DougBurke, line 1028:
```suggestion
                raise IOError("Error generating WMAP w/sky2tdet!") from exc
```

bin/specextract
```diff
@@ -783,14 +812,14 @@ def pi_alternate_wmap(pi,wmap):
     dmcopy.punlearn()
     dmcopy.infile = tmp_wmap.name
     dmcopy.outfile = pi
-    dmcopy.clobber = "yes"
-    dmcopy.verbose = "0"
+    dmcopy.clobber = True
+    dmcopy.verbose = 0
     dmcopy()
     
     dmappend.punlearn()
```
DougBurke, line 819: Has this been tested? It doesn't look dmappend has been imported - e.g. it's not in

```python
from ciao_contrib.runtool import dmextract, mkrmf, mkacisrmf, mkarf, mkwarf, arfcorr, asphist, dmgroup, dmhedit, dmcopy, acis_fef_lookup, sky2tdet, combine_spectra, ardlib, acis_set_ardlib, dmmakereg, add_tool_history, dmhistory, dmimgthresh
```

bin/specextract
```diff
@@ -429,18 +428,35 @@ def determine_rmf_tool(infile,rmffile):
...
+    instrument = args["instrument"]
+    phafile = args["phafile"]
+    evtfile = args["filename"].split("/")[-1]
+    
+    if instrument == "ACIS":
+        rmffile_ccd = args["rmffile_ccd"]
+        rmftool = _determine_rmf_tool(phafile,rmffile_ccd,evtfile)
+
+        return {key_id : {"rmftool" : rmftool}}
+
```
DougBurke, line 450: This should handle the not-an-ACIS instrument case. Either erroring out or returning `None` or `{}` or something.

I'd also suggest handing this as a special case, so you have

```python
    if instrument != "ACIS":
        return-something-or-error-out

    rmffile_ccd = args["rmffile_ccd"]
    rmftool = _determine_rmf_tool(phafile,rmffile_ccd,evtfile)
    return {key_id : {"rmftool" : rmftool}}
```

bin/specextract
```diff
@@ -341,8 +347,8 @@ def create_arf_ext(full_outroot, ebin,
 
         mkwarf()
 			
-    except Exception as err_msg:
-        print(err_msg)
+    except Exception as exc:
+        print(exc)
 
         raise IOError(f"Failure to create weighted ARF.  Possible causes include: zero counts in the input region; a memory allocation error; or corrupt ARDLIB.  Try running {toolname} with weight=no, binarfmap!=1, or punlearn ardlib.")
```
DougBurke, line 353:
```suggestion
        raise IOError(f"Failure to create weighted ARF.  Possible causes include: zero counts in the input region; a memory allocation error; or corrupt ARDLIB.  Try running {toolname} with weight=no, binarfmap!=1, or punlearn ardlib.") from exc
```

ciao_contrib/_tools/specextract.py
```diff
@@ -535,7 +541,7 @@ def _check_region_crlf(self,inf):
                         if b"\r\n" in r.read():
                             ## n.b.: modify region file by doing
                             ## `cat regfile | tr -d '\015' > regfile.lf-only`
-                            raise IOError(f"EOL: the line ends in the ASCII region file, '{regfile}', use Windows/DOS-style carriage return & line feed (CRLF) rather than Unix-compatible line feed only.")
+                            raise IOError(f"EOL: the line ends in the ASCII region file, '{regfile}', using Windows/DOS-style carriage return & line feed (CRLF) rather than Unix-compatible line feed only.")
 
                 except FileNotFoundError:
                     raise IOError(f"region file '{regfile}' not found")
```
DougBurke, line 547:
```suggestion
                    raise IOError(f"region file '{regfile}' not found") from None
```

ciao_contrib/_tools/specextract.py
```diff
@@ -447,7 +453,7 @@ def _event_stats(self,file,colname,args=None):
 
             with new_pfiles_environemt(ardlib=False,copyuser=False):
```
DougBurke, line 454:
```suggestion
            with new_pfiles_environment(ardlib=False,copyuser=False):
```

This suggests we don't have code that tests this section, as it should have crashed and burned

bin/specextract
nplee: this has in the distant past, but is unused, since the time savings is minimal with parallelization.  I've pulled it.

ciao_contrib/_tools/specextract.py
nplee: fixed, but reviewing the code a bit more, this conditional use case for the function is not called on in the script

bin/specextract
nplee: revised in latest commit

PR #684, opened 2022-11-15
srcflux: add aplimits plugin and update Changes

kglotfelty:
This is a replacement for #675 

Instead of putting the `aplimits` code directly into `srcflux`, this provides a sample plugin so that users who want to use it can run it that way.

### DIFF
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -6,7 +6,7 @@ team at PSU. These are labeled below. Please see
 
   https://cxc.cfa.harvard.edu/ciao/guides/srcflux_for_ae_users.html
 
-for more information on using srcflux as a replacement for
+for more information on using srcflux as a substitute for
 acis_extract.
 
 The ahelp files for the contributed scripts have been re-organized,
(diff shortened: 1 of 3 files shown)


data/sample_srcflux_plugins.py
```diff
@@ -366,3 +366,48 @@ def spectral_fit_sample_srcflux_obsid_plugin(infile, outroot, band, elo, ehi, sr
 
     return retval
     
+
+
+def aplimits_srcflux_obsid_plugin(infile, outroot, band, elo, ehi, src_num):
+    """
+    Sample plugin: aplimits
+    
+    This plugin runs the new `aplimits` script to compute
```
DougBurke, line 375:
```suggestion
    This plugin runs the `aplimits` script to compute
```

Very substantial comment, I know.

Changes.CIAO_scripts
```diff
@@ -127,6 +127,22 @@ Updated scripts
      - Extract radial profiles (per obsid)
      - Perform blind deconvolution using arestore (per obsid)
      - Estimate source extent (per obsid)
+     - Get upper limits on missed detections and count rates for false
+       detections using the new aplimits tools (per obsid)
```
DougBurke, line 131: I'm okay wit the `new` here :-)

Issue #686, opened 2022-11-17
specextract: the combined output name?


DougBurke:
Testing #676 I noticed

```
% specextract @infiles"[sky=region(src.reg)]" combine+
Warning:  found possibly out of date user parameter file, renamed to /home/dburke/cxcds_param4/specextract_20221117.10:25:12.par
Output directory path + root name for output files (): foo/bar
Running specextract
Version: 15 November 2022

Extracting Spectra:[████████████████████████████████████████████████████████████] 4/4

Generating Aspect Histograms & Weights Maps:[████████████████████████████████████████████████████████████] 4/4

Generating ARFs & RMFs:[████████████████████████████████████████████████████████████] 8/8

Grouping Spectra:[████████████████████████████████████████████████████████████] 4/4

Adding Header Keywords:[████████████████████████████████████████████████████████████] 4/4

```

Looking in the output directory I'd expect them all to begin "bar" but I get

```
% ls -1 foo/
ba_combined_src.arf
ba_combined_src.pi
ba_combined_src.rmf
bar_src1.arf
bar_src1_grp.pi
bar_src1.pi
bar_src1.rmf
bar_src2.arf
bar_src2_grp.pi
bar_src2.pi
bar_src2.rmf
bar_src3.arf
bar_src3_grp.pi
bar_src3.pi
bar_src3.rmf
bar_src4.arf
bar_src4_grp.pi
bar_src4.pi
bar_src4.rmf
```

My guess is the logic to create the "combined" file name is, in this particular case, removing one character too many (ie it's changed "bar" to "ba").

DougBurke:
This is also present in CIAO 4.14

DougBurke:
Why are computers?

```
>>> orig = "foo/bar_src1"
>>> orig
'foo/bar_src1'
>>> orig.rstrip("_src1")
'foo/ba'
```

but

```
>>> "xsaaa".rstrip("aaa")
'xs'
>>> "xs_aaa".rstrip("aaa")
'xs_'
>>> "xs_aaa".rstrip("_aaa")
'xs'
```

nplee:
ok, it looks like rstrip is not functioning like I was expecting it to.  I've fixed it.

DougBurke:
Aha: turns out it's because of how `rstrip` is meant to work: the argument is not the actual text to remove, but the list of white space characters to remove. So if the `outroot` parameter ends in one of the characters from "src1" then it gets removed. e.g.

```
[...]

PR #687, opened 2022-11-17
specextract: truncated outroot when combine=yes

nplee:
When combine=yes, the outroot of the combined file is erroneously truncated by one character due to unexpected behavior of rstrip with underscores on string variables.  If the rstrip is applied directly to a string, the behavior is as expected.

### DIFF
--- a/bin/specextract
+++ b/bin/specextract
@@ -1292,20 +1292,19 @@ def spectra(args):
 
         try:
             phafile = extract_spectra(full_outroot,fullfile,ptype,channel,
-                                      ewmap,binwmap,instrument,clobber,verbose)
-
-            ## update extracted readout streak spectrum effective exposure time ##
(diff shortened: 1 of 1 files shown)


DougBurke:
Can you check #688 - and the explanation in #686 - as `rstrip("src1")` is going to work but that's only because the preceeding character is "_", which stops the `rstrip` call from removing more characters. The better solution is just to remove the number of characters - e.g. "foo[:-5]", which I did in #688.

nplee:
@DougBurke looks good.  I've also moved around around the exception call for extract_spectra, that you had mentioned in the earlier PR review, in the latest commit. The regtests show that change didn't cause a problem, if it's worth updating right now too.

DougBurke:
Replaced by #688

Issue #696, opened 2022-11-22
specextract with mask regions


kglotfelty:
This is a regression in 4.15 -- but not at all a show stopper.  From the DM pixelmask test suite, `specextract` used to work with `sky=mask()`, but now it fails

```bash
$ CT_INDIR=/data/sdsreg/current_regtest/input/dm_pixelmask/
$ evt=$CT_INDIR/4425/repro/acisf04425_repro_evt2.fits
$ img=$CT_INDIR/4425/fluximage/test_broad_thresh.img

$ dmimgcalc $img none ones op="imgout=(img1-img1)+1" clob+
$ dmcopy "ones[sky=circle(4108,4045,8)][opt full]" ones_circle.mask clob+

$ specextract ${evt}"[sky=mask(ones_circle.mask)]" se_out clob+ bkgresp- weight=no 
Running specextract
Version: 18 November 2022

# specextract (18 November 2022): ERROR 'mask'
```

Not sure what's changed here.  Just a note that the `pyregion` module does not know/understand `mask` .

Using a regular, simple regions works:

```bash
$ specextract ${evt}"[sky=circle(4108,4045,8)]" se_out clob+ bkgresp- weight=no 
Running specextract
Version: 18 November 2022

Extracting Spectra:[#######################################################] 1/1

Generating Aspect Histograms & Weights Maps:[##############################] 1/1

Generating ARFs & RMFs:[###################################################] 2/2

Grouping Spectra:[#########################################################] 1/1

Adding Header Keywords:[###################################################] 1/1
```

nplee:
This is emerging from the usage of the "resp_pos" parameter.  If resp_pos=MAX (as in the past) or CENTROID is used, it runs to completion with a mask.

kglotfelty:
Ahh ... great.  Thanks, that works.

A better error message might be better here, but probably not worth it (this is more academic than an actual use case).

PR #700, opened 2022-11-29
specextract: raise exception when pixel mask is incompatible with 'resp_pos'

nplee:
This addresses #696, and raises an exception if a pixel mask is used instead of a region if `resp_pos=MAX` or `resp_pos=CENTROID` since `pyregion` does not support pixel masks.  There may be some clever workaround, but for now, this suffices.

Code also linted and cleaned up.

### DIFF
--- a/ciao_contrib/_tools/specextract.py
+++ b/ciao_contrib/_tools/specextract.py
@@ -25,7 +25,7 @@ Routines to support the specextract tool.
 
 """
 
-import os, sys, numpy, cxcdm, paramio, stk, tempfile, region
+import os, sys, numpy, cxcdm, paramio, stk, region
 import pycrates as pcr
 
 from ciao_contrib.runtool import dmstat, dmcoords, dmlist, acis_fef_lookup, get_sky_limits, new_pfiles_environment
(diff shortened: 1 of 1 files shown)


ciao_contrib/_tools/specextract.py
```diff
@@ -1507,7 +1508,7 @@ def obs_args(self,src_stk,out_stk,bkg_stk,asp_stk,bpixfile_stk,rmffile,
                             v1("Found a Level=3 ahst3.fits file in 'asp' input; the 'asphist' block corresponding to the source region location will be used.\n")
 
                         if stk_count["asol"] != 1:
-                            asphist_arg = f"{stk_dict['asol'][i]}[asphist{self._event_stats(fullfile,'ccd_id')}]"
+                            asphist_arg = f"{asp_stk[i]}[asphist{self._event_stats(fullfile,'ccd_id')}]"
```
DougBurke, line 1511: So is this change a bug fix?

nplee: yes; weirdly, it's never been triggered in regression testing or anyone reporting this and was only caught by pylint.  Likely wasn't caught since I've yet to see anyone use aspect histograms as input except for our tests....

PR #705, opened 2022-12-08
Improve aplimits

hamogu:
I analyzed a failure in the regression tests that I wrote.
I relaized that the difference is entirely due to numerics, so this PR increases the number of stps taken in the numerical integration. 

Since I'm at it and I installed a new spell checking tool in my editor recently, I also fixed a few speling mistakes and added more docs to the code to help me next time I'm forced to think about each step again.

### DIFF
--- a/bin/aplimits
+++ b/bin/aplimits
@@ -289,8 +289,9 @@ class APLimitsMarginalize(APLimits):
         # width = sqrt(variance)
         # -10 * std to + 20 * std
         # It's asymmetric because the function is for small numbers.
-        # choice of 10, 20 and 100 points is made by looking at functions
-        # plotted for typical values.
+        # choice of 1000 points is made by looking at functions
+        # for typical input values and checking how many point we need to get
+        # reasonable precision.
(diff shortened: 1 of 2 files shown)


bin/aplimits
```diff
@@ -265,7 +291,7 @@ class APLimitsMarginalize(APLimits):
         # It's asymmetric because the function is for small numbers.
         # choice of 10, 20 and 100 points is made by looking at functions
```
DougBurke, line 292: Is this comment out of date now, or is it unrelated to the size of x (i.e. the fact we now use 1000 rather than 100)?

bin/aplimits
```diff
@@ -265,7 +291,7 @@ class APLimitsMarginalize(APLimits):
...
         # plotted for typical values.
```
hamogu, line 293:
```suggestion
        # choice of 1000 points is made by looking at functions
        # for typical input values and checking how many point we need to get
        # reasonable precision.
```

bin/aplimits
hamogu: Comment updated.

bin/aplimits
DougBurke: Ta

PR #707, opened 2022-12-09
combine_spectra: Fix origin of exposure time

hamogu:
The exposure time was taken from the ARF, even when exp_origin='pha' was set (the default),
simple because an ARF exits.
In Chandra, we probably never noticed, but I run into a problem when applying
this script to NICER spectra, where the ARF does not have all the header
information that the PHA would have.
The bug is somewhat obvious in the code and can be tested with any ARF file that does not have an EXPOSURE keyword.
One would get
```
OSError: The file 'NICER_SOMETHING.pi' is missing the 'EXPOSURE' keyword.
```
However, the usual wrapping of the errors in scripts makes it somewhat harder to get to see the details of the error.

### DIFF
--- a/bin/combine_spectra
+++ b/bin/combine_spectra
@@ -19,7 +19,7 @@
 #
 
 toolname = "combine_spectra"
-__revision__ = "07 April 2020"
+__revision__ = "09 December 2022"
 
 import sys
 import os
@@ -106,8 +106,8 @@ class Spectrum():
             self.arffile = self._get_filename_if_not_none( infile,tab, "ancrfile", pc.read_file)
         else:
             #make sure can read it
-            tab=pc.read_file( self.arffile )
-            if not tab.column_exists("specresp"):
+            arf = pc.read_file( self.arffile )
+            if not arf.column_exists("specresp"):
                 raise IOError("File '{}' does not appear to be an ARF".format(self.arffile))
 
         if exppref == "pha" or is_none(self.arffile):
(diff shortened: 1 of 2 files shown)


share/doc/xml/combine_spectra.xml
```diff
@@ -361,6 +361,13 @@ SOURCE_BACKSCAL = (src_exp1*src_backscal1) + ... + (src_expN*src_backscalN) / (s
        </LIST>
 </ADESC>
 
+  <ADESC title="Changes in the script 4.15.1 (April 2023) release">
```
hamogu, line 364: Here I'm looking into the future. How do we usually mark that up for an unreleased version with no date yet?

bin/combine_spectra
```diff
@@ -106,7 +106,7 @@ class Spectrum():
             self.arffile = self._get_filename_if_not_none( infile,tab, "ancrfile", pc.read_file)
         else:
             #make sure can read it
-            tab=pc.read_file( self.arffile )
+            pc.read_file( self.arffile )
```
hamogu, line 109: look a few lines down where we check if the rmf file is readable without overwriting the variable `tab` which later is assumed to be the pha file.

hamogu:
Not a show-stopper for 4.15. It's a coincidence that I hit that now, because I've already released marx and now I'm doing some science again.

share/doc/xml/combine_spectra.xml
DougBurke: Well, sometimes we don't bother putting something in and hope we remember to add it in before a release. We sometimes use placeholder text (e.g. "XXX" rather than a month name). We plan to release something this month so this time you can use January ;-)

hamogu, line 364:
```suggestion
  <ADESC title="Changes in the script 4.15.1 (January 2023) release">
```

hamogu:
Release date updated.

Issue #703, opened 2022-12-05
name resolver fails badly


kglotfelty:
I'm not sure why we are seeing this now (maybe something with the fall-through to the secondary name server).

This is happening with 4.14 and 4.15 on mac (Monterey), ciao-install editions (so python 3.8 and 3.9), x86.

```bash
% find_chandra_obsid "FOOBAR" 
# find_chandra_obsid (27 January 2022): ERROR 'str' object has no attribute 'errno'


% find_chandra_obsid "FOOBAR" --debug verbose=5
Running: find_chandra_obsid
  version: 27 January 2022
Querying name resolvers for name=FOOBAR
CADC name query: https://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/cadc-target-resolver/find?format=ascii&service=all&cached=true&target=FOOBAR
Error opening URL: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1125)>
error.reason = [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1125)
error.reason.errno = 1
Skipping to an unverified SSL context
Unable to query with no context!
Error opening URL: HTTP Error 425: 
error.reason = 
# find_chandra_obsid (27 January 2022): ERROR 'str' object has no attribute 'errno'


##### Traceback:

Traceback (most recent call last):
  File "/Applications/ciao-4.14/ots/lib/python3.8/urllib/request.py", line 1354, in do_open
    h.request(req.get_method(), req.selector, req.data, headers,
  File "/Applications/ciao-4.14/ots/lib/python3.8/http/client.py", line 1255, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/Applications/ciao-4.14/ots/lib/python3.8/http/client.py", line 1301, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/Applications/ciao-4.14/ots/lib/python3.8/http/client.py", line 1250, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/Applications/ciao-4.14/ots/lib/python3.8/http/client.py", line 1010, in _send_output
    self.send(msg)
  File "/Applications/ciao-4.14/ots/lib/python3.8/http/client.py", line 950, in send
    self.connect()
[...]

PR #719, opened 2023-01-05
Improve URL access (hopefully)

DougBurke:
So, if the SSL library doesn't work for your system then we end up trying an "unverified" SSL context in the URL code. Unfortunately we did not handle that call in the same way we did to the original urlopen call, which could lead to the odd behavior we saw in #703. Even if it didn't, this version should be *slightly* better.

### DIFF
--- a/coords/resolver.py
+++ b/coords/resolver.py
@@ -1,5 +1,5 @@
 #
-#  Copyright (C) 2011, 2013, 2015, 2016, 2018, 2019, 2020, 2022
+#  Copyright (C) 2011, 2013, 2015, 2016, 2018, 2019, 2020, 2022, 2023
 #  Smithsonian Astrophysical Observatory
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -85,10 +85,10 @@ def strtofloat(val):
         return float(val)
 
     except ValueError:
-        raise ValueError("Unable to convert '{0}' to a float.".format(val))
+        raise ValueError(f"Unable to convert '{val}' to a float.") from None
 
 
-def query_url(name, url, method):
+def query_url(name, url, method, context=None):
     """Wrap up handling of errors for a URL call.
 
     Parameters
@@ -96,6 +96,8 @@ def query_url(name, url, method):
     name : str
     url : str
     method : {"CADC", "Sesame"}
+    context : optional
+       If set, send to urlopen
 
     """
 
@@ -104,8 +106,8 @@ def query_url(name, url, method):
     # different.
     #
     try:
-        v5(f"{method} name query: {url}")
-        return urlopen(url)
+        v5(f"{method} name query: {url}    context is None: {context is None}")
+        return urlopen(url, context=context)
 
     except HTTPError as he:
         # HTTPError is a subclass of URLError so process first
@@ -113,7 +115,7 @@ def query_url(name, url, method):
         v5(f"Error from {method} name resolver - status = {code}")
 
         if method == "CADC" and code == 425:
-            raise ValueError(f"No position found matching the name '{name}'.")
+            raise ValueError(f"No position found matching the name '{name}'.") from he
 
         v5(str(he))
         raise he
@@ -129,24 +131,18 @@ def query_url(name, url, method):
         # there a better one?
         #
         reason = str(ue.reason)
-        if reason.find("CERTIFICATE_VERIFY_FAILED") != -1:
-            # Try with no context
+        if context is None and reason.find("CERTIFICATE_VERIFY_FAILED") != -1:
+            # Try with an unverified context - which we pass back to query_url so
+            # that we do not have to repeat all the try/except work.
             #
             v5("Skipping to an unverified SSL context")
             context = ssl._create_unverified_context()
-            try:
-                return urlopen(url, context=context)
-            except URLError as ue2:
-                v5("Unable to query with no context!")
-                v5(f"Error opening URL: {ue2}")
-                v5(f"error.reason = {ue2.reason}")
-                v5(f"error.reason.errno = {ue2.reason.errno}")
-                raise ue2
+            return query_url(name, url, method, context=context)
 
         if ue.reason.errno == 8:
-            raise IOError(f"Unable to connect to the {method} Name Resolver")
+            raise IOError(f"Unable to connect to the {method} Name Resolver") from ue
 
-        raise
+        raise ue
 
 
 def identify_name_cadc(name):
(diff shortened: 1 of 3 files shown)


DougBurke:
@kglotfelty - I don't know if this could have been the reason for #703 (I can come up with a reason that this new behavior is better / would not show the issue) but it's hard to be sure. Does this change make sense?

DougBurke:
On my home ubuntu 22.10 laptop with a ciao-install version I get, before this fix

```
% find_chandra_obsid neptune v=5
Running: find_chandra_obsid
  version: 27 January 2022
Querying name resolvers for name=neptune
CADC name query: https://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/cadc-target-resolver/find?format=ascii&service=all&cached=true&target=neptune
Error opening URL: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1129)>
error.reason = [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1129)
error.reason.errno = 1
Skipping to an unverified SSL context
Unable to query with no context!
Error opening URL: HTTP Error 425: 
error.reason = 
# find_chandra_obsid (27 January 2022): ERROR 'str' object has no attribute 'errno'
```

but after

```
% find_chandra_obsid neptune v=5
Running: find_chandra_obsid
  version: 27 January 2022
Querying name resolvers for name=neptune
CADC name query: https://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/cadc-target-resolver/find?format=ascii&service=all&cached=true&target=neptune    context is None: True
Error opening URL: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1129)>
error.reason = [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1129)
error.reason.errno = 1
Skipping to an unverified SSL context
CADC name query: https://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/cadc-target-resolver/find?format=ascii&service=all&cached=true&target=neptune    context is None: False
Error from CADC name resolver - status = 425
Sesame name query: https://cdsweb.u-strasbg.fr/cgi-bin/nph-sesame/-ox/~SNV?neptune    context is None: True
Error opening URL: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1129)>
error.reason = [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1129)
error.reason.errno = 1
Skipping to an unverified SSL context
Sesame name query: https://cdsweb.u-strasbg.fr/cgi-bin/nph-sesame/-ox/~SNV?neptune    context is None: False
Response from query:
<?xml version="1.0" encoding="UTF-8"?>
<Sesame xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 xsi:noNamespaceSchemaLocation="http://vizier.u-strasbg.fr/xml/sesame_4x.xsd">
<Target option="~SNV">
  <name>neptune</name>
  <!-- Q37165161 #1 -->
  <INFO> *** NNNothing found *** </INFO>
</Target>
</Sesame>
<!--- ====Done (2023-Jan-06,03:22:06z)==== -->

Parsing response as XML
# find_chandra_obsid (27 January 2022): ERROR No position found matching the name 'neptune'.
```

or, without the verbosity

```
% find_chandra_obsid neptune
# find_chandra_obsid (27 January 2022): ERROR No position found matching the name 'neptune'.
```

so it seems to be doing the right thing.

coords/resolver.py
```diff
@@ -104,16 +106,16 @@ def query_url(name, url, method):
     # different.
     #
     try:
-        v5(f"{method} name query: {url}")
-        return urlopen(url)
+        v5(f"{method} name query: {url}    context is None: {context is None}")
```
kglotfelty, line 109: I assume that `{context is None`} works w/ f strigs, but it does read funny.   As long as it works, should be ok though.

coords/resolver.py
```diff
@@ -85,17 +85,19 @@ def strtofloat(val):
         return float(val)
 
     except ValueError:
-        raise ValueError("Unable to convert '{0}' to a float.".format(val))
+        raise ValueError(f"Unable to convert '{val}' to a float.") from None
```
kglotfelty, line 88: Showing my python ignorance ... is `raise stuff from None` any different than just `raise stuff` ?  If so then okay, if not then a little confusing.

kglotfelty (approved):
i haven't actually run it but beside a couple of comments but probably ok.

coords/resolver.py
DougBurke: Yeah. It works

```
>>> context = None
>>> f"Is context None: {context is None}"
'Is context None: True'
>>> context = True
>>> f"Is context None: {context is None}"
'Is context None: False'
```

coords/resolver.py
DougBurke: You can 

- raise Blah()
- raise Blah() from None
- raise Blah() from some_previous_exception

What the latter do is change what happens if you have code that has caught exceptions: the None case means something like "hey, don't bother reporting the other errors" whereas giving a previous execption means "the traceback will identify as these are related". The following overly-complicated example hopefully helps:

# using "from None"

Note we get a nice clean trace here, and this is what we want here (we are really just replacing the error with a more-informative one).

```
>>> try:
...     try:
...         raise ValueError("x")
...     except ValueError:
...         raise RuntimeError("bloop") from None
... except ValueError:
...     print("pass")
... 
Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
RuntimeError: bloop
```

# using no "from ..."

Here we see both exceptions but they appear "un-related" - i.e. awkward to track down

```
>>> try:
...     try:
...         raise ValueError("x")
...     except ValueError:
...         raise RuntimeError("bloop")
... except ValueError:
...     print("pass")
... 
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
ValueError: x

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
RuntimeError: bloop
```

# with "from ..."

It's still wordy, but at least we know there's some relation between the two exceptions (which can be useful when you want to know what the previous exception was, as is the case in some of the "oh, we failed because of a SSL exception, let's trey again but turning off the SSL context checks" type code we have.

```
>>> try:
...     try:
...         raise ValueError("x")
...     except ValueError as ve:
...         raise RuntimeError("bloop") from ve
... except ValueError:
...     print("pass")
... 
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
ValueError: x

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
RuntimeError: bloop
```

HTH

PR #722, opened 2023-01-06
specextract fails when blanksky/stowed background files with OBS_ID keyword

nplee:
script fixed helpdesk ticket#024432 which for some reason was failing when the stowed background file being used to extract the background file from was failing due to having an OBS_ID keyword.  Reason still unclear, but an unused temporary file was somehow being moved to a different memory location and could not be closed successfully which threw a `FileNotFoundError`.  While passing the exception allowed the script to run successfully, the better solution was to re-work things slightly and avoid creating the unused temporary file to begin with when background responses are not being created.

Additional linting applied and changed discarded `weightfile` to use `tempfile` context manager.

Testing also revealed that script history wasn't being written correctly to output files when run directly from the command-line, although non-TTY regression tests wrote the history correctly.  Bug traced to failing to close parameter file during a check when setting verbosity.

### DIFF
--- a/bin/specextract
+++ b/bin/specextract
@@ -34,7 +34,7 @@ Script:
 __version__ = "CIAO 4.15"
 
 __toolname__ = "specextract"
-__revision__ = "5 January 2023"
+__revision__ = "9 January 2023"
 
 
 ##########################################################################
@@ -55,7 +55,8 @@ from ciao_contrib.logger_wrapper import initialize_logger, make_verbose_level, s
 from ciao_contrib.param_wrapper import open_param_file
 from ciao_contrib.cxcdm_wrapper import get_info_from_file
 
-from ciao_contrib.runtool import dmextract, mkrmf, mkacisrmf, mkarf, mkwarf, arfcorr, asphist, dmgroup, dmhedit, dmcopy, acis_fef_lookup, sky2tdet, combine_spectra, ardlib, acis_set_ardlib, dmmakereg, add_tool_history, dmhistory, dmimgthresh
+from ciao_contrib.runtool import dmextract, mkrmf, mkacisrmf, mkarf, mkwarf, arfcorr, dmgroup, dmhedit, dmcopy, acis_fef_lookup, sky2tdet, combine_spectra, ardlib, acis_set_ardlib, dmmakereg, add_tool_history, dmhistory, dmimgthresh
+from ciao_contrib.runtool import asphist as asphist_tool
 
 import ciao_contrib._tools.fileio as fileio
 import ciao_contrib._tools.utils as utils
@@ -171,7 +172,7 @@ def readout_streak_spec(specfile,infile,src_x,src_y):
                     D = 2.0 * r.radii.max()
 
             except (AttributeError,TypeError) as exc:
-                raise IOError("There is an unexpected issue with the readout streak extrction region type.") from exc
+                raise IOError("There is an unexpected issue with the readout streak extraction region type.") from exc
 
             nrows.append(numpy.ceil(D) * include_reg)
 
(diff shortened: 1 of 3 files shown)


bin/specextract
```diff
@@ -1648,24 +1633,32 @@ def resps(args,specdicts,specextract_args,pars,parallelize):
...
+
+    tdetwmap = {d["key_id"] :
+                None if all([d["srcbkg"] == "bkg", not d["dobkgresp"]])
+                else tempfile.NamedTemporaryFile(suffix="_tdet",dir=d["tmpdir"])
+                for d in args}
 
-    tdetwmap = {d["key_id"] : tempfile.NamedTemporaryFile(suffix="_tdet",dir=d["tmpdir"]) for d in args}
-    tdetwmap_name = {k:v.name for k,v in tdetwmap.items()}
+    asphists_name = {k:v.name if v is not None else v for k,v in asphists.items()}
+    tdetwmap_name = {k:v.name if v is not None else v for k,v in tdetwmap.items()}
```
DougBurke, line 1647: I haven't followed the logic through below, so maybe the suggestion doesn't make sense, but can you drop those elements where `v is None` from the dictionaries here?

DougBurke:
@nplee - can you also update the ahelp file here?

bin/specextract
nplee: I'll double check. I'm pretty sure that's what I originally tried, but things were going a bit wonky downstream in some regression tests without being so explicit here or I'd need to fall back on using some try/except statements.

DougBurke: There cuold be a downstream reason why it's useful to keep the `None` values

nplee:
@DougBurke I've updated the ahelp, did some more linting removing unused declared variables, renamed using the `asphist` tool since there's too many instances of variables using the same name as the tool, and fixed a bug where HRC background responses are created even when `bkgresp=no` in the latest commit.

bin/specextract
```diff
@@ -722,17 +723,16 @@ def set_badpix(evtfile, bpixfile, instrument, verbose):
             raise IOError(f"Failed to set {bpixfile} bad pixel file in ardlib.par for {evtfile}.") from exc
 
     else:
-        bpixpar = f"AXAF_{fileio.get_keys_from_file(evtfile)['DETNAM']}_BADPIX_FILE"
+        detnam = fileio.get_keys_from_file(evtfile)["DETNAM"]
```
DougBurke, line 726:
```suggestion
        detnam = fileio.get_keys_from_file(evtfile)["DETNAM"].replace("-", "_")
```

and then below you can replace `bpixvar.replace(...)` with `bpixvar` (I think there's two of them; I haven't checked closely to know if it breaks anything).

share/doc/xml/specextract.xml
```diff
@@ -1869,7 +1869,7 @@ unix% cat output.lis
 	region with 'resp_pos=REGION'.
       </PARA>
       <PARA title="Parallel processing">
-	The way that the tools are run in parallel has been reworked
+	The way that the tools are run in parallel has beenx reworked
```
DougBurke, line 1872:
```suggestion
	The way that the tools are run in parallel has been reworked
```

share/doc/xml/specextract.xml
```diff
@@ -1891,6 +1891,17 @@ unix% cat output.lis
...
+	When using a blanksky or stowed background file for background
+	spectrum extraction where the OBS_ID keyword is included would
+	fail with 4.15.0 which has been re-worked.
+      </PARA>
+      <PARA>
+	Also, a bug where an incomplete specextract history written to
+	the output files when using TTY screen output and a bug where
+	HRC background responses were created even with 'bkgresp=no'
+	have been fixed.
+      </PARA>
```
DougBurke, line 1904: How about, for these two paragraphs, we replace with something like

```
<PARA>
A number of bugs added in 4.15.0 have been fixed: the script history is now written out correctly when run intereactively,  the presence of the OBS_ID keyword in blanksky or stowed background files is no-longer a problem, and the script no-longer creates a HRC background response when 'bkgresp=no' is used.
</PARA>
```

(I think the HRC issue is technically not a 4.15.0 problem but not worth separating out, I think).

PR #728, opened 2023-01-13
Allow XSPEC FORTAN models that use udmget to be built

DougBurke:
This is highly experimental, as I don't have a model that actually uses the udmget code to check it works.


nplee:
@DougBurke check helpdesk#022079 I believe has a Fortran-based model that uses udmget attached.

DougBurke:
@nplee - thanks. This seemed to have worked (at least, things didn't start falling out of the sky)...

PR #729, opened 2023-01-14
Update runtool for color_color

DougBurke:


DougBurke:
@nplee - do you want to take this (it is not the world's biggest problem if we don't take it)

PR #737, opened 2023-02-03
get_dither_parameters: estimate dither parameters from asol file

kglotfelty:
new script to estimate the dither parameters:  amplitude and period, from the aspect solution file.    

The values are estimates/approximates only.

We don't try to estimate the phase (probably doable but not very interesting).

Needs lots of testing especially with
- short observations , ~1ksec to check if powerspectrum works
- observation w/o dither (eg Crab)
- calibrations "Big" dither observations where Y and Z values are very different
- any other non-standard dither configurations


kglotfelty:
need to up pfold search to 3500 (obs 6158 uses 3333).

PR #747, opened 2023-03-20
tweak to support libraries on MacOS

nplee:
On MacOS, `sysconfig.get_config_var(SHLIB_SUFFIX)` returns `.so` even though pretty much all shared libraries have the `.dylib` extension.  As is, the script will error out on any MacOS system before getting to try any linkers set in `LDFLAGS`.

### DIFF
--- a/bin/convert_xspec_user_model
+++ b/bin/convert_xspec_user_model
@@ -67,7 +67,7 @@ for a description of the model.dat format.
 """
 
 toolname = "convert_xspec_user_model"
-toolver  = "13 January 2023"
+toolver  = "20 March 2023"
 
 import sys
 import os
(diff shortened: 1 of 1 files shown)


DougBurke:
I'm fairly sure at one point I found a setting I could use to tell me if it's .s o or .dylib, but could never find it again...

nplee:
> I'm fairly sure at one point I found a setting I could use to tell me if it's .s o or .dylib, but could never find it again...

I did some more poking around after I made the PR... could it be `sysconfig.get_config_var("WITH_DYLD")` that you're thinking about?

At least on my RHEL workstation, it returns `0` while on my Mac it returns `1`.

PR #759, opened 2023-04-24
energy_hue_map

kglotfelty:
new script that takes mean energy map and counts to create a true color image


DougBurke:
So, I think the following is at the level of "not sure the complexity is worth it", but I wonder if we could add a "smooth each band" argument - i.e. something that would essentially run `crates_contrib.utils.smooth_image_crate` on each band (use the same smoothing per band). There's enough complexity here that it may not be worth it (and `smooth_image_crate` isn't really the API needed by this script), but thought I'd ask (as someone who often goes straight to ds9's true-color menu and applies the same smoothing to each band).

Or maybe a companion script to smooth all images in a crate with the same options.

Or just forget about it...

kglotfelty:
For the testing I've been doing, I've been smoothing and exposure correcting the input "counts" image (`infile`) and it works well.

kglotfelty:
I also find doing a little smoothing (just a 3x3 boxcar) to the input mean energy map to "round" off the sharp edges looks better.

There's definitely an argument not to do, but we're ignoring PSF blurring when claiming independence of the different map regions so I think that gives a little room for "ad hoc"/"artistic" license here.

PR #766, opened 2023-05-05
convert_xspec_script: initial support for MDEFINE

DougBurke:
The idea was to add support for the MDEFINE keyword. It got a bit out of hand:

- Allow model expressions to refer to compiled user models; this requires that `convert_xspec_user_model modname <model.dat>` was used to process the code, and then `--models modname` is added to the `convert_xspec_script` call
- the parsing and evaluation of the model language has been reworked and will hopefully remove some edge cases (e.g. random extra "( * " statements and some cases where an extra space is removed), but I can not guarantee that this hasn't just broken something somewhere else. We can still have too many brackets, but it should not affect the results...
- support for MDEFINE
  - this should at least allow any valid MDEFINE line to be read in
  - it doesn't guarantee that the model will wok
    - convolution models are explicitly known not to work
    - models that call XSPEC functions (e.g. HEAVISIDE) should work but it's likely at least one function doesn't match XSPECs behavior
    - everything else is "it should work but  the user should review the results", in part because of differences between XSPEC and Python languages
- there's some internal changes related to filtering; the most obvious is that notice/ignore calls should now generate the same logging output that the ui.notice/ignore calls create (namely telling you how the filter has changed).
  - there's also an attempt to support some files which can have different groups when read in by XSPEC and Sherpa, in a "XSPEC appears to be silently correcting an invalid file" case but in such a way that we don't have an exact mapping between the channel ranges. If you have a file like this you should now get a warning (whereas before the script would have failed, so if no-one asked about this then they can't have hit this situation!)
  - there's some slight changes to the output script - we now create a warning call rather than a boring print statement in some cases

I am not particularly happy with the ergonomics or name of the `--models` option, but not sure I can be bothered to change it. Any suggestions?

For my reference: the [MDEFINE](https://heasarc.gsfc.nasa.gov/docs/xanadu/xspec/manual/XSmdefine.html) command.

### DIFF
--- a/bin/convert_xspec_script
+++ b/bin/convert_xspec_script
@@ -95,7 +95,7 @@ def check_clobber(fname):
 
 
 @lw.handle_ciao_errors(TOOLNAME, TOOLVER)
-def convert(infile, outfile, clean=False, clobber=False):
+def convert(infile, outfile, models=None, clean=False, clobber=False):
 
     if outfile != '-' and not clobber:
         check_clobber(outfile)
@@ -108,7 +108,7 @@ def convert(infile, outfile, clean=False, clobber=False):
         except OSError as oe:
             raise OSError(f"Unable to read from '{infile}'") from oe
 
-    out = xcm.convert(cts, clean=clean)
+    out = xcm.convert(cts, models=models, clean=clean)
     if outfile == '-':
         sys.stdout.write(out)
     else:
@@ -123,6 +123,9 @@ if __name__ == "__main__":
 
     parser.add_argument("infile", help="XCM file to convert (- for stdin)")
     parser.add_argument("outfile", help="Sherpa file to create (- for stdout)")
+
+    parser.add_argument("--models", action="append",
+                        help="Add models created by convert_xspec_user_model: modulename")
 
     parser.add_argument("--clean", dest="clean", action="store_true",
                         default=False,
@@ -161,5 +164,6 @@ if __name__ == "__main__":
         sys.exit(0)
 
     convert(args.infile, args.outfile,
+            models=args.models,
             clean=args.clean,
             clobber=args.clobber)
--- a/bin/convert_xspec_script
+++ b/bin/convert_xspec_script
@@ -2,7 +2,7 @@
 
 #
 # Copyright (C) 2020, 2021, 2023
-#           Smithsonian Astrophysical Observatory
+# Smithsonian Astrophysical Observatory
 #
 #
 # This program is free software; you can redistribute it and/or modify
@@ -52,7 +52,7 @@ from sherpa_contrib.xspec import xcm
 
 
 TOOLNAME = "convert_xspec_script"
-TOOLVER = "12 January 2023"
+TOOLVER = "05 May 2023"
 
 lw.initialize_logger(TOOLNAME, verbose=1)
 V1 = lw.make_verbose_level(TOOLNAME, 1)
@@ -68,7 +68,7 @@ The conversion is not guaranteed to match XSPEC perfectly.
 
 COPYRIGHT_STR = """
 Copyright (C) 2020, 2021, 2023
-          Smithsonian Astrophysical Observatory
+Smithsonian Astrophysical Observatory
 
 This program is free software; you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
--- a/bin/convert_xspec_script
+++ b/bin/convert_xspec_script
@@ -52,7 +52,7 @@ from sherpa_contrib.xspec import xcm
 
 
 TOOLNAME = "convert_xspec_script"
-TOOLVER = "08 May 2023"
+TOOLVER = "09 May 2023"
 
 lw.initialize_logger(TOOLNAME, verbose=1)
 V1 = lw.make_verbose_level(TOOLNAME, 1)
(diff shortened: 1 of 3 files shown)


DougBurke:
There's at least one stupid mistake in this

```
< set_source(1, m1 * m2 * ( * (m3) * m4 + m5 * (m6)  + m7 * m8(m9)))
---
> set_source(1, m1 * m2 * ( * (m3( * m4 + m5 * (m6(  + m7 * m8(m9)()
```

Although the model expression was not exactly that great in the original...

DougBurke:
@nplee I think I'm bout done adding functionality, so if you can test this out that would be great.

DougBurke:
Here's one example where the model-parsing has changed, but not fixed, things:

Released:

```
% echo -e "model (phabs)(apec)" | convert_xspec_script - - --verbose 0
from sherpa.astro.ui import *


# model (phabs)(apec)
m1 = create_model_component('xsphabs', 'm1')
m2 = create_model_component('xsapec', 'm2')

# Set up the model expressions
#
set_source(1, (m1) *  * (m2))
```

Now:

```
% echo -e "model (phabs)(apec)" | convert_xspec_script - - --verbose 0
from sherpa.astro.ui import *


# model (phabs)(apec)
m1 = create_model_component('xsphabs', 'm1')
m2 = create_model_component('xsapec', 'm2')

# Parameter settings

# Set up the model expressions
#
set_source(1, (m1)(m2))
```

DougBurke:
This s definitely something that changed. I thought I had tested it but obviously not. Ta for the heads up.

nplee:
In `xcm.py`:
https://github.com/cxcsds/ciao-contrib/blob/380eaa7d70bcbf1b93d9fa134571c371b18aa9c9/sherpa_contrib/xspec/xcm.py#L1927

should be changed to `if command == 'backgrnd':` to get explicit background files to be picked up.

DougBurke:
@nplee - this should "fix" things. 

```
% echo "model constant*phabs( ( zpowerlw )etable{mytorus_Ezero_v00.fits} + constant( atable{mytorus_scatteredH500_v00.fits} ) + constant*gsmooth( atable{mytl_V000010nEp000H500_v00.fits} ) )" | convert_xspec_script - - --verbose=0
from sherpa.astro.ui import *


# model constant*phabs( ( zpowerlw )etable{mytorus_Ezero_v00.fits} + constant( atable{mytorus_scatteredH500_v00.fits} ) + constant*gsmooth( atable{mytl_V000010nEp000H500_v00.fits} ) )
m1 = create_model_component('xsconstant', 'm1')
m2 = create_model_component('xsphabs', 'm2')
m3 = create_model_component('xszpowerlw', 'm3')
load_xstable_model('m4', 'mytorus_Ezero_v00.fits', etable=True)
m4 = get_model_component('m4')
m5 = create_model_component('xsconstant', 'm5')
load_xstable_model('m6', 'mytorus_scatteredH500_v00.fits')
m6 = get_model_component('m6')
m7 = create_model_component('xsconstant', 'm7')
m8 = create_model_component('xsgsmooth', 'm8')
load_xstable_model('m9', 'mytl_V000010nEp000H500_v00.fits')
m9 = get_model_component('m9')

# Parameter settings

# Set up the model expressions
#
set_source(1, m1 * m2 * ((m3) * m4 + m5 * (m6) + m7 * m8((m9))))
```

Note that the source model is not great - the `()` around m3 and m6 are not ideal, the one around `m9` looks more serious but it seems to be okay (at least from my limited testing)

DougBurke:
This is an example of using Python typing to test / check the code. Whether it's a successful or useful experiment is left for the future to decide.

DougBurke:
There's the possibility of adding a 'import numpy as np' line to the output, as this would be needed if a user-model uses functions like `ATAN`: if you load with `%run foo.py` then it would be needed, but if you use `%run -i foo.py` from within Sherpa it would not be needed.

nplee:
@DougBurke in `xcm.py`, `load_background` should be changed to `load_bkg`.

https://github.com/cxcsds/ciao-contrib/blob/8b6fc2cc0b031fe2a0c52f7bff9082bbbe1ce0a9/sherpa_contrib/xspec/xcm.py#L1012
and
https://github.com/cxcsds/ciao-contrib/blob/8b6fc2cc0b031fe2a0c52f7bff9082bbbe1ce0a9/sherpa_contrib/xspec/xcm.py#L1023

nplee:
My mistake, those tweaks aren't complete, since it fails to also assign a dataset ID to the background... c.f.

```
load_pha(1, '5005_xaf_82_grp.pi', use_errors=True)
load_bkg('5005_xaf_82_bkg.pi')
load_pha(2, '15186_xaf_82_grp.pi', use_errors=True)
load_bkg('15186_xaf_82_bkg.pi')
load_pha(3, '16564_xaf_82_grp.pi', use_errors=True)
load_bkg('16564_xaf_82_bkg.pi')
load_pha(4, '16565_xaf_82_grp.pi', use_errors=True)
load_bkg('16565_xaf_82_bkg.pi')
```

DougBurke:
Bah. Humbug.

DougBurke:
I've added fixes for 

- https://github.com/cxcsds/ciao-contrib/pull/766#issuecomment-1548755131 
- @nplee 's background issue

and hopefully haven't broken anything in the process

PR #767, opened 2023-05-16
lightcurves.py: use Agg backend is $DISPLAY not set

nplee:
switch matplotlib backend to Agg to avoid any warning being thrown if in `lc_clean` or `lc_sigma_clip` are with `plot=True` if `$DISPLAY` is not set.


DougBurke:
Ta.

PR #771, opened 2023-06-12
add energies command to convert_xspec_script

DougBurke:
Add initial support for the [ENERGIES](https://heasarc.gsfc.nasa.gov/xanadu/xspec/manual/XSenergies.html) command from XSPEC.

- not tested
- does not cover all use cases
- the parsing is likely incomplete


DougBurke:
Dang it - it looks like we need https://github.com/sherpa/sherpa/pull/1617 (at least) before this will work in CIAO, which would imply this is not for CIAO 4.15 (unless we monkeypatch things, I guess, but it's not likely to be a nice monkey to play with)

PR #776, opened 2023-07-11
marx_rl_decon: new script to perform Richardson Lucy deconvolution ...

kglotfelty:
…using marx to perform the PSF convolutions.

This needs the bug fix in https://github.com/kglotfelty/marx/tree/fix_offset_input_image to work.

### DIFF
--- a/bin/marx_rl_decon
+++ b/bin/marx_rl_decon
@@ -1,0 +1,307 @@
+#!/usr/bin/env python
+# coding: utf-8
+#
+# Copyright (C) 2023 Smithsonian Astrophysical Observatory
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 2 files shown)


DougBurke:
Do we think we want to wait until marx is released with this (so go out with it post the December release), or include it now?

kglotfelty:
I realized that there is a normalization problem somewhere -- I think I know where but I haven't had time to go about fixing.  [Basically I believe I need to filter/clip by the FOV file when combining multi-obi, but it has to happen at the right place.]  

I'm doubtful it'll be ready for 4.16.0

DougBurke:
So, I think it's safest to not hurry for the 4.16.0 release then.

hamogu:
The marx source IMAGE also does not work as it is used here, that's not just a simple bug, it's designed to work in a different way (see discussion in https://github.com/Chandra-MARX/marx/issues/50). I think what's needed is a new marx source ("IMAGE-WCS") which makes full use of the WCS information: https://github.com/Chandra-MARX/marx/issues/57

PR #795, opened 2023-09-08
Add datatype parameter to apply_fov_limits (fix #793)

### LINKED ISSUE #793: apply_fov_limits when counts overflow short integer
Not sure if this only applies to `apply_fov_limits` but it's using the default short-integer data-type when binning.  This can lead to overflow for certain datasets, eg:

```bash
$ apply_fov_limits infile="acisf04396N004_evt2.fits[energy=500:7000]" outfile="acisf04396_broad_thresh.img" binsize="4" fovfile="acisf04396_000N004_fov1.fits" tmpdir="/pool7" clobber="yes" verbose="1" 
Running: apply_fov_limits
  version: 01 April 2020
Observation: ObsId 4396 - ACIS-01236
Using ccd_id=0,1,2,3,6 from acisf04396N004_evt2.fits[energy=500:7000]
Using FOV file acisf04396_000N004_fov1.fits.

The output image will have 1027 by 739 pixels, pixel size of 1.968 arcsec,
    and cover x=1480.5:5588.5:4,y=2552.5:5508.5:4.

Created: acisf04396_broad_thresh.img

$ dmstat acisf04396_broad_thresh.img cen-
EVENTS_IMAGE
    min:	-31645 	      @:	( 4118.5 4162.5 )
    max:	28033 	      @:	( 4118.5 4154.5 )
   mean:	1.6392095426 
  sigma:	79.528731832 
    sum:	1244083 
   good:	758953 
   null:	0 
```

Negative counts are never a good thing.

```bash
$ dmcopy infile="acisf04396N004_evt2.fits[energy=500:7000][bin x=1480.5:5588.5:4,y=2552.5:5508.5:4]" outfile="acisf04396_broad_thresh.img" kernel="default" option="" verbose="1" clobber="yes" 
The debug level is 1 
CXC Data Model: DM V2.10
[*1]  Binning will overflow i2 datatype, try [opt type=i4]
# dmcopy (CIAO 4.15): Binning will overflow i2 datatype, try [opt type=i4]
$ echo $?
0
```
Though of course `dmcopy` is not particularly helpful here.

Probably need to hard-code to use `[opt type=i4]`

DougBurke:
The script now defaults to using i4 for the output image type, but it can be switched back to i2 using the new datatype parameter.

There are also a number of format to f-string changes.

Although technically a simple change, given that this

- adds a parameter
- the behavior has been like this a long time
- changes a bunch of other unrelated code because I can
- has not been tested

I'm labeling this for 4.16 rather than any rush to get it into 4.15.


DougBurke:
@kglotfelty - no rush needed for a review

kglotfelty (approved):
Changes look ok.  I would have just hard coded to `i4` rather than add a new parameter but that's just a choice.  Tested and looks good.  

Aside:  interesting that the old N003 version of obsid 4396 has a max of 32206 events in a pixel so doesn't hit the 32768 rollover; the new N004 version of the data hits a max of 33891.

DougBurke:
Ta for checking

PR #806, opened 2023-09-27
Srcflux new regions

kglotfelty:
This adds a new `regions` parameter to allow users to select how the default regions are created.

`simple` is the current method using circular ECFs and remains the default.

`optimized` is the acis-extract-like way which uses polygon ECFs for src region and backgrounds with a fixed number of counts.


kglotfelty:
ok to get into ciaox for more testing/feedback

share/doc/xml/srcflux.xml
```diff
@@ -929,6 +1044,49 @@ Summary of merged source fluxes
...
       </PARAM>
 
+    <PARAM name="regions" type="string" reqd="no" def="simple">
+      <SYNOPSIS>Algorithm used to create default source and background regions</SYNOPSIS>
+      <DESC>
+        <PARA>
+          If the srcreg and bkgreg parameters are left blank, then
+          srcflux will automatically create source and background regions;
+          this is done separately for each observation.  There are two
+          options:
```
DougBurke, line 1054: There's actually three options: what about "`user`"?

kglotfelty: added user

DougBurke: Ta

DougBurke:
Can you add something to the changelog too (I'm  just busy doing other things and also my magic GH token has died and I'm too lazy too magic up a new one ;-)

kglotfelty:
:heavy_check_mark:

PR #811, opened 2023-10-04
Dax model editor check

kglotfelty:
Adds a property to each entry box widget to identify when the value has been modified but not yet set (press `Return`/`Enter`).

Checks that all entry box widgets have been set before allowing user to `Fit`, `Plot`, run `Conf`, or `Quit` -- `Quit` resumes the sherpa fitting script so we want to be sure that the values set in the GUI are what sherpa is actually using.

### DIFF
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -2,7 +2,14 @@
 
 TODO: overview of changes.
 
+
 Using CIAO tools from ds9 (dax)
+
+  The model editor now checks that any values that have been modified
+  have been set (by pressing the Return or Enter key) before allowing
+  the user to proceed with certain actions such as Fit or Plot.
+  Modified values that have not been set appear as red text; the text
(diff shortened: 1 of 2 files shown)


DougBurke:
What's a good entry in the changelog for this. I'm thinking of the following but is "The model editor" the right term (and am I missing anything else?)

```
Using CIAO tools from ds9 (dax)

    The model editor now requires that values must be explicitly changed (with Return or Enter) before
    an action - such as Fit or Plot - is selected.

```

kglotfelty:
How about:


```text
The model editor now checks that any values that have been modified have been 
set (by pressing the Return or Enter key) before allowing the user to proceed with
certain actions such as Fit or Plot.  Modified values that have not been set appear
as red text; the text changes to black after the values have been set.
```

DougBurke:
LGTM

PR #817, opened 2023-10-20
Make pget less chatty

DougBurke:
This makes the output of `./mk_runtool.py` match the previous version (before I changed it to use paramio to query the parameter file).

### DIFF
--- a/mk_runtool.py
+++ b/mk_runtool.py
@@ -44,11 +44,12 @@ be handled more sensibly.
 """
 
 import glob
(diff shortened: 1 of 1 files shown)


kglotfelty (approved):
Looks OK ... but is `ERR_BUFFER_OPEN` needed?  It just looks like you set the value but don't use it anywhere.

DougBurke:
I am just future proofing myself for a future that will never happen.

PR #818, opened 2023-10-24
reproject_obs fails if mask files cannot be found

nplee:
from a helpdesk ticket, we've found that `reproject_obs` requires the presence of mask files in the working directory by its initial file validation checks even though they are not used by the script.  This update fixes the problem should mask files are unavailable.

### DIFF
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -54,6 +54,12 @@ Updated scripts
     The script has been updated to support changes to Sherpa plotting
     in this release.
 
+  reproject_obs
+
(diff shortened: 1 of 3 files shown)


DougBurke:
So, this is the sort of thing we could have in a 4.15 contrib release (if we decide we want one).

nplee:
@DougBurke Yes, the `merge_obs` regtests on 4.15 all passed with this tweak.

nplee:
whoops, well, that was fun fixing up commits after forgetting to rebase first... but the change log and ahelp have been updated.  Hopefully it's mergeable now!

PR #821, opened 2023-11-02
add support for L3 asolfile w/ CONTENT=ASPSOL3 and evt3 w/o TITLE keyword

### LINKED ISSUE #769: Support for L3 data products?
A helpdesk user, Ref: 24690, tried to run `srcflux` using the L3 data products (`bpix3`, `asol3`, and `regevt3`).  It didn't go well.  All the results were `NaN` and `fluximage` failed with

```bash
# fluximage (04 November 2021) : ERROR Aspect solutions have CONTEXT=ASPSOL or ASPSOLOBI, but found ASPSOL3 in pcadf23022_000N020_asol3.fits.gz
```
Should we support obi-level Level 3 files with the contrib scripts?  No chance of supporting stk based products.

Purist would say maybe not since the L3 calibrations are probably different than what user has in CIAO now; but users do use L2 straight from the archive w/o running `chandra_repro` so there's already some mismatch.

kglotfelty:
This updates the low level i/o routines to accept `CONTENT="ASPSOL3"` to be the same as `ASPSOL`.  (upshot: need to run `skyfov method=minmax`) .

It also relaxes `specextract` requirement for keys to check for merged files since `evt3` file omits `TITLE`.  

This will close #769

### DIFF
--- a/ciao_contrib/_tools/specextract.py
+++ b/ciao_contrib/_tools/specextract.py
@@ -892,25 +892,9 @@ class ParDicts(object):
         """
 
         keys_to_check = ['TITLE', 'OBSERVER', 'OBJECT', 'OBS_ID', 'DS_IDENT']
-        merge_key = []
         for mkey in keys_to_check:
-            if mkey in kwdict:
-                merge_key.append(kwdict[mkey].lower())
-
-        # ~ merge_key = [kwdict["TITLE"].lower(),
-                     # ~ kwdict["OBSERVER"].lower(),
-                     # ~ kwdict["OBJECT"].lower(),
-                     # ~ kwdict["OBS_ID"].lower()]
-
-        # ~ try:
-            # ~ merge_key.append(kwdict["DS_IDENT"].lower())
-        # ~ except KeyError:
-            # ~ pass
-
-        if "merged" in merge_key:
-            raise IOError(f"Merged data sets are unsupported by {toolname}.  Merged events files should not be used for spectral analysis.") # or v1 warning?
-
-        del(merge_key)
+            if mkey in kwdict and kwdict[mkey].lower() == "merged":
+                raise IOError(f"Merged data sets are unsupported by {toolname}.  Merged events files should not be used for spectral analysis.") # or v1 warning?
 
 
     def _check_blanksky(self,headerkeys,srcbkg_kw,bkgresp):
--- a/ciao_contrib/_tools/specextract.py
+++ b/ciao_contrib/_tools/specextract.py
@@ -188,10 +188,10 @@ class ParDicts(object):
         src_cnt = len(src_stk)
         asptype = self._get_keyvals(asp_stk, "CONTENT")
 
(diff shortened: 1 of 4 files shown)


ciao_contrib/_tools/specextract.py
```diff
@@ -891,15 +891,21 @@ def _check_merged_input(self,kwdict,toolname):
...
+
+        # ~ merge_key = [kwdict["TITLE"].lower(),
+                     # ~ kwdict["OBSERVER"].lower(),
+                     # ~ kwdict["OBJECT"].lower(),
+                     # ~ kwdict["OBS_ID"].lower()]
+
+        # ~ try:
+            # ~ merge_key.append(kwdict["DS_IDENT"].lower())
+        # ~ except KeyError:
+            # ~ pass
```
DougBurke, line 908:
```suggestion
```

ciao_contrib/_tools/specextract.py
```diff
@@ -891,15 +891,21 @@ def _check_merged_input(self,kwdict,toolname):
...
-
-        try:
-            merge_key.append(kwdict["DS_IDENT"].lower())
-        except KeyError:
-            pass
+        keys_to_check = ['TITLE', 'OBSERVER', 'OBJECT', 'OBS_ID', 'DS_IDENT']
+        merge_key = []
+        for mkey in keys_to_check:
+            if mkey in kwdict:
+                merge_key.append(kwdict[mkey].lower())
```
DougBurke, line 898: So, I think all we care about in this routine is finding a value of `"merged". If so we can just drop the `merge_key` array and do something like (following not checked)

```
for key in keys_to_check:
    try:
        val = kwdict[key]
    except KeyError:
        continue

    if val.lower() == "merged":
        raise OSError("....")
````

nplee (approved):
These look good to me.
I'll merge/tweak the specextract changes in the version I've been working on my local branch as well.

DougBurke:
Can we have updates for changelog and probably ahelp files too now the code side is settled?

kglotfelty:
Well, I don't want to say you can, even hypothetically, start using L3 files -- because you shouldn't ... so don't really want to open that door for people to possibly miss use stuff.  So is something generic like

> Added support for additional variants of standard data products

sufficiently accurate and obscure that it'll be the no-op it's intended to be?

DougBurke:
I had thought at least some change was generic, but I guess generic in the sense of "allows it to get further with a merged product, whether this is from the catalog, merge_obs, or ..." so it's all still under the "don't ask, don't merge" policy.

kglotfelty:
Sorry, to clarify ... the per-obi `evt3` file (not the merge `stkevt3`) omits the `TITLE` keyword.  No idea why -- it should be the same as the L1/L2 ... no merging going on, but that's for another day (my head still hurts too much from last week to pick at this scab).    The missing `TITLE` keyword then caused `specextract` to raise a `KeyError` exception; this fixes that.

Technically it is generic ... any event file that is missing the `TITLE` keyword will no longer trigger this exception -- but the only specimen in the wild we have is from csc.

PR #824, opened 2023-11-09
blanksky* revisions for contrib 4.16.0

nplee:
`blanksky`, `blanksky_image`, are `blanksky_sample` have undergone linting and clean up.

`blanksky` will make use of the `CHKVFPHA` keyword introduced by `acis_process_events` in CIAO 4.16, but the fragile history parsing approach to determine if `status=0` event filtering for ACIS VFaint data is also retained for reference event files haven't been reprocessed using 4.16.

`blanksky_sample` also will replace the `Merged`-valued keywords in the `psf_bkg_out` and `fill_out` files files, which combine source `infile` events and background events via `dmmerge` with more helpful information using the keyword values from the original `infile`.

### DIFF
--- a/bin/blanksky
+++ b/bin/blanksky
@@ -19,7 +19,7 @@
 #
 
 __toolname__ = "blanksky"
-__revision__  = "31 October 2023"
+__revision__  = "07 November 2023"
 
 
 import os
@@ -224,6 +224,15 @@ def check_tool_par(evt,tool,param):
     par_instance = [par.replace("\"","") for par in par_instance]
 
     return f"{param}=yes" in par_instance
+
+
+
+def _chkvfpha(kw,evt):
+    if kw["DATAMODE"] == "VFAINT":
+        if "CHKVFPHA" in kw.keys():
+            return kw["CHKVFPHA"] # kw introduced in 4.16 to replace fragile 'check_tool_par' approach
+        return check_tool_par(evt=evt,tool="acis_process_events", param="check_vf_pha")
+    return False
 
 
 
@@ -816,12 +825,7 @@ is on ACIS-{aimpoint}, due to exceeding of SIM-Z limit.\n")
 
         # if datamode is VFAINT and ran a_p_e with check_vf_pha=yes,
         # remove background events with non-0 status
-        if False not in (keywords["DATAMODE"] == "VFAINT",
-                         check_tool_par(evt=infile,tool="acis_process_events",
-                                        param="check_vf_pha")):
-        # in CIAO 4.16, replace with:
-        # if keywords["DATAMODE"] == "VFAINT" and keyword["CHKVFPHA"]:
-        # and can remove 'check_tool_par' function
+        if keywords["DATAMODE"] == "VFAINT" and _chkvfpha(keywords,infile):
             dmcopy.punlearn()
 
             dmcopy.infile = f"{bkg_calib.name}[status=0]"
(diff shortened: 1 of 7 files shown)


bin/blanksky
```diff
@@ -162,28 +172,35 @@ def e_norm(evtfile,bkgfile,chip,bkgparams,instrument):
...
     obs_counts = float(dmlist())
 
     dmlist.punlearn()
-    dmlist.infile = bkgfile+ccd+bkgparams
+    dmlist.infile = f"{bkgfile}{ccd}{bkgparams}"
     dmlist.opt = "counts"
 
     bkg_counts = float(dmlist())
 
     return obs_counts/bkg_counts
```
DougBurke, line 186: Not for this release, but I think it would be easier to uderstand if we just used crates and did something like (untested)

    obs_counts = read_file(f"{evtfile}{ccd}{bkgparams").get_nrows()

and something similar for `bkg_counts`

bin/blanksky
```diff
@@ -375,53 +401,55 @@ def apply_gain(infile,outfile,kw,tmpdir,verbose):
...
+
+        a_p_e.gainfile = f"{aciscaldb}/det_gain/{gainfile}"
+        a_p_e.ctifile = f"{aciscaldb}/t_gain/{tgainfile}"
+        a_p_e.tgainfile = f"{aciscaldb}/cti/{ctifile}"
 
         try:
-            acis_process_events.infile = infile
-            acis_process_events()
+            a_p_e.infile = infile
+            a_p_e()
```
DougBurke, line 434: you should just be able to say

    a_p_e(infile=infile)

here but again not an important change.

bin/blanksky
nplee: 👍I was actually considering this during the clean up but wasn't sure if it's a preferable approach...

bin/blanksky_sample
```diff
@@ -100,67 +100,74 @@ def get_par(argv):
 
     ## error out if there are spaces in absolute paths of various parameters
     if " " in os.path.abspath(params["infile"]):
```
DougBurke, line 102: Not for this PR, but it looks like the "check for spaces" check could be made into a routine, something like

```
def check_no_spaces(infile, filetype):
    abspath = os.path.abspath(infile)
    if " " not in abspath:
        return

    return OSError(f"The asolute path for the {filetype} file, {abspath}, cannot contain any spaces")
```

and then called with something like

```
    for inkey, filetype in [("infile", "source evemts"), ("bkgfile", "blanksky background"), ...]:
        check_no_spaces(params[inkey], filetype)
```

[although I haven't gone through to see if this handles all cases]

DougBurke:
You have - both in the ahelp file and the changelog, comments like `file has been linted and cleaned up` and `no change to the UX`. For 99.99% of the users they don't care/won't understand what this means. So can we drop these lines?

nplee:
yes, absolutely.  Do you want me to push and update or do you want to interactively make the tweak while merging?

share/doc/xml/blanksky.xml
```diff
@@ -343,7 +343,7 @@
 
     <ADESC title="Changes in the November 2023 revision (scripts 4.16.0)">
       <PARA>
-	Use the <tt>CHKVFPHA</tt> header keyword, introduced via <HREF
+	Use the CHKVFPHA header keyword, introduced via <HREF
```
DougBurke, line 346: good spot

PR #825, opened 2023-11-10
Update energy_hue_map.par

kglotfelty:
fix dump typo in par file


kglotfelty:
We will need to update the `runtool` module, **but**, I don't want to right now because CIAOT has all those extra tools that shouldn't be there.

PR #826, opened 2023-11-13
specextract 4.16.0 MARX event file support and 'none' ancillary file

nplee:
Code linting and support for MARX-projected event files added.  Better handling of ancillary files set to `none` added, in particular for when no bad pixel files are used, prior behavior could result in the `ardlib` falling back on the CALDB bad pixel file instead.


ciao_contrib/_tools/specextract.py
```diff
@@ -909,6 +909,17 @@ def _check_merged_input(self,kwdict,toolname):
...
+        """
+        check if the input file has been generated by psf_project_ray
+        """
+        keys_to_check = [key for key in kwdict.keys() if key.upper().startswith("HDUCLAS")]
+
+        for pkey in keys_to_check:
+            if kwdict["HDUNAME"].lower() == "rayevent" and kwdict[pkey].lower() == "psfray":
+                raise IOError(f"Event files generated by psf_project_ray are not supported by {toolname}.")
+
+
```
DougBurke, line 922: So, if I'm reading this correctly then we cam have

```
def _check_ppr(sef, kwdict, toolname):
    if kwdict["HDUNAME"].lower() != "rayevent":
        return

    keys_to_check = [...]
    for pkey in keys_to_check:
        if kwdict[pkey].lower() == "psfray":
            raise IOError(...)
```

and actually we can just combine the two loops - ie don't create `keys_to_check` so we just have

````
    for key, val in kwdict.items():
        if not key.upper().startswith("HDUCLAS") or val.lower() != "psfray":
            continue
        raise IOError(...)
```

nplee: I've updated the function and tested it out... looks good

Issue #819, opened 2023-10-26
review version coded numpy links


kglotfelty:
In several ahelp files we have links to specific version of numpy (1.7.0).  We should review and use a version-independent link if possible (me thinks, since no idea what version users will get w/ conda)

- cda_search.xml
- imextent.xml
- make_image_crate.xml
- make_table_crate.xml
- write_columns.xml

PR #828, opened 2023-11-13
Update links

DougBurke:


DougBurke:
@kglotfelty - this should fix #819 but am I missing anything?

PR #839, opened 2023-11-17
added `xpaset` region formatting specification

nplee:
The `dax` regression tests were failing on a system where the DS9 preference file has region file formatting set to `ciao` rather than the default `ds9`.  The two shell scripts have been modified so that the `xpaset` region calls also specifies the formatting.

The regression tests subsequently run without any further changes on RHEL8 or modifying/deleting the preference file. 

However, on MacOS it seems that the regression test will persist to fail unless `bash` is upgraded to a version > 3 and at this time, the regression test itself may need modification to also set the region formatting for the `xpaset` calls... I'm still exploring this.

### DIFF
--- a/bin/ds9_snap2.sh
+++ b/bin/ds9_snap2.sh
@@ -56,7 +56,7 @@ fi
 new=`pget dmstat out_cntrd_phys`
 
 xpaset -p $ds9 regions delete all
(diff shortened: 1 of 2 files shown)


DougBurke:
Since this is a cross-generational issue I've asked for a review by @kglotfelty

kglotfelty (approved):
Tested and looks good.

Issue #843, opened 2023-12-09
specextract: Extracting regions with 0 counts


hamogu:
When I run specextract in a region that has zero counts, I get https://github.com/cxcsds/ciao-contrib/blob/eaba8ccfe4c00a7b7387e826245ce7d5fd5c7443/ciao_contrib/_tools/specextract.py#L446

The error suggests that the region has to be specificed in sky pixel coordinates. That's not true - I can use any of the CIAO regions (e.g. RA/DEC) and I'll get the same extraction with the same number of counts (0) and thus the same error.

Should we downgrade this from an error to a warning? I get that in most cases it probably means that a user put in the wrong coordinates and for that a hard stop is good. However, there are some edge cases where it might be useful (and, you guessed it, I'm trying one of those, hence I'm hitting this limit) to continue running and get an ARF and RMF even if there are 0 counts: The ARF and RMF can be used in simulations with e.g. Sherpa to fake certain models and determine how likely it is that some model gives 0 counts. Of course, usually one would take a simpler approach, but particularly when looking at a source that dithers on and off the chip or that is very soft or very hard (so that an "average PSF" or exposure map created for a specific energy is not good enough) running spectral models with ARF/RMF should be more accurate. 
However, I acknowledge that this is an unusual use of specexctract, so maybe it's better to serve the general population of users for more common use cases better and keep the hard error?

nplee:
@hamogu I can certainly improve the message, but for your specific example case, where the extraction region may have zero counts, also define the `refcoord` parameter (https://cxc.cfa.harvard.edu/ciao/ahelp/specextract.html#plist.refcoord) to force the script to move on from the counts check... I think it may be limited to unweighted responses though.

hamogu:
See #844 for a suggestion for the error message.

I saw that refcoord reference, but then it was Saturday morning and I decided to do something else. I might also make a local copy of the script without the check or just call "mkacisrmf" etc separately by hand.

DougBurke:
@nplee do we want the fix in #844 ?

nplee:
@DougBurke I think the update to the error messages are a perfectly reasonable solution to the issue.

PR #844, opened 2023-12-09
Generalize message in specextract

hamogu:
This addresses the first point in #843.

### DIFF
--- a/ciao_contrib/_tools/specextract.py
+++ b/ciao_contrib/_tools/specextract.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2021-2023
+#  Copyright (C) 2021, 2023
 #        Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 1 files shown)


ciao_contrib/_tools/specextract.py
```diff
@@ -443,7 +443,7 @@ def _check_event_stats(self,file,refcoord_check=None,
         if counts == "0":
             try:
                 if refcoord_check.lower() in ["","none","indef"]:
-                    raise IOError(f"{file} has zero counts. Check that the region format is in sky pixels coordinates.")
+                    raise IOError(f"{file} has zero counts. Check that the region correct (e.g. wrong region, coordinates not in sky pixels or degrees).")
```
hamogu, line 446: Do we want to put the workaround to run this anyway in the message or do we thing it's better to leave that for expert users to figure out themselves on the thought that in most cases zero counts actually indicates an error in the region?

```suggestion
                    raise IOError(f"{file} has zero counts. Check that the region correct (e.g. wrong region, coordinates not in sky pixels or degrees) or use parameter "refcoord"")
```

DougBurke: Let's be explicit...

DougBurke, line 446:
```suggestion
                    raise IOError(f"{file} has zero counts. Check that the region correct (e.g. wrong region, coordinates not in sky pixels or degrees) or use parameter='refcoord'.")
```

DougBurke: Redid the suggestion to fix a typo (use of `"` rather than `'`)

PR #845, opened 2023-12-10
Usability in aplimits

hamogu:
aplimits can take either bkg_rate or m (number of counts in background region) by setting one of the two to INDEF.
However, m needed to be an int, not INDEF before this PR.
That's not a big problem, because bkg_rate takes precedence (if set, any int number of m will do), but unexpected (no setting for m should be required).
While I'm at it, also fix a few typos in docstrings.

### DIFF
--- a/bin/aplimits
+++ b/bin/aplimits
@@ -332,8 +332,8 @@ def aplimits(prob_false_detection=.1, prob_missed_detection=.5, T_s=1, A_s=1,
     parameters prob_false_detection and prob_missed_detection, while Kashyap et al. (2010) call
     them alpha and beta, which might be confusing because aplimits uses
     alpha and beta for different concepts. Similarly, Kashyap et al. (2010)
-    only have an error ratio "r", while in CIAO we measure source and
-    background areas.
+    only have an area ratio "r", while in CIAO we measure source and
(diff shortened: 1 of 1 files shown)


DougBurke:
Should the parameter file change from 

```
m,i,a,,0,,"Number of counts in background aperture"
```

to

```
m,i,a,0,0,,"Number of counts in background aperture"
```

just to make it clear. This may be a question for @kglotfelty - since I get

```
>>> import paramio
>>> paramio.pget("aplimits", "m")
'0'
```

which suggests the default value is 0 here.

kglotfelty:
I haven't read the docs so not really sure the impact of the change, but it might be you want the default to be `INDEF`?

```
m,i,a,INDEF,0,,"Number of counts in background aperture"
```
Indeed a blank value for a numeric datatype (`r` or `i` is treated at `0` instead of `INDEF`)

DougBurke:
As there's been no progress on the parameter setting I'm just going to merge as is.

DougBurke:
I don't think it needs an entry in the changelog, unless @hamogu thinks otherwise (it is not clear to me if this fixes a user error or is just a nice-to-have change).

PR #851, opened 2024-01-13
map2reg: combine regions using CXCRegion instead of dumping to ascii file

kglotfelty:
Unfortunately need to bite the bullet and combine regions using more memory intensive way.  Reading ASCII region with many sided polygon fails.

### DIFF
--- a/bin/map2reg
+++ b/bin/map2reg
@@ -1,4 +1,4 @@
-#!/soft/miniconda/envs/contrib_test/bin/python3.11
+#!/usr/bin/env python
 #
 # Copyright (C) 2019, 2024 Smithsonian Astrophysical Observatory
 #
--- a/bin/map2reg
+++ b/bin/map2reg
@@ -1,6 +1,6 @@
-#!/usr/bin/env python
+#!/soft/miniconda/envs/contrib_test/bin/python3.11
 #
-# Copyright (C) 2019, 2023 Smithsonian Astrophysical Observatory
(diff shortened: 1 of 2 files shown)


bin/map2reg
```diff
@@ -1,6 +1,6 @@
-#!/usr/bin/env python
+#!/soft/miniconda/envs/contrib_test/bin/python3.11
```
DougBurke, line 1: We don't want this change

bin/map2reg
```diff
@@ -136,20 +140,12 @@ def main():
...
-    # an ASCII region file and then reading back in as a single
-    # CXCRegion is much faster
-
-    # Combine into single region file
-    tmpfile = NamedTemporaryFile(dir=os.environ["ASCDS_WORK_PATH"],
-                                 suffix=".reg", delete=True)
-    with open(tmpfile.name, "w") as outfile:
-        for myreg in outreg:
-            outfile.write(myreg+"\n")
+    # Unfortunately writing all out to an ascii file sometimes fails
```
DougBurke, line 143: I assume there's no easy (or maybe any) way to tell when the dump to an ASCII file approach will work?

kglotfelty: There's probably a per-line string length defined somewhere in the DM (it parses the `region()` syntax and sends lines to region lib).  Probably some variation of n*1024 +/- m depending on how new-line/string terminator are handled.

bin/map2reg
kglotfelty: fixed

PR #852, opened 2024-01-13
Merge too small fix

### LINKED ISSUE #849: merge_too_small failes to merge mapIDs with 0 counts
The error is related to how the tool checks if a mapID is present in input file. This works for what was the original `method=area` option, but for `method=counts` this means that mapIDs with 0 counts are not merged with other mapIDs to meet the minimum requirement.

https://github.com/cxcsds/ciao-contrib/blob/a83799bb9b9bf5d223c95ae18f764950d17c1386/bin/merge_too_small#L85

https://github.com/cxcsds/ciao-contrib/blob/a83799bb9b9bf5d223c95ae18f764950d17c1386/bin/merge_too_small#L105

Fix is kinda easy, but initial testing is running far too slow so may need to rework logic to avoid some duplication of effort.

kglotfelty:
This fixes #849 .

Also improves speed, and fixes problem when diagonal neighbors would lead to disjoint polygons (polygons that don't share a side, just a corner).

### DIFF
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -1,3 +1,14 @@
+## 4.16.1 - February 2024
+
+Updated scripts
+
+  merge_too_small
+
+    When using method=counts, it no longer skips regions with 0 counts.
+    Pixels must share an edge to be considered neighbors (so diagonally
+    adjacent pixels are no-longer considered touching).
+
+
 ## 4.16.0 - December 2023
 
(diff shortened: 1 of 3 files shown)


DougBurke:
Does the changelog change make sense? I wasn't sure of how much faster the rework is (is it always, or only sometimes) so didn't mention it, but we can do so.

kglotfelty:
changelog looks good.  except for small images (10's x 10's) it should be faster, but I don't think worth mentioning.

PR #853, opened 2024-01-14
centroid_map: fix issue with NaNs in input image

### LINKED ISSUE #841: centroid map with NaNs
the `centroid_map` script is not correctly ignoring pixel values set to NaN's when computing the centroid.  Not a show-stopper.

kglotfelty:
This closes #841

### DIFF
--- a/bin/centroid_map
+++ b/bin/centroid_map
@@ -25,9 +25,10 @@ import os
 import numpy as np
 import ciao_contrib.logger_wrapper as lw
 from pycrates import read_file
+from crates_contrib.masked_image_crate import MaskedIMAGECrate
 
 __TOOLNAME__ = "centroid_map"
-__REVISION__ = "24 August 2023"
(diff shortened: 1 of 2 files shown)


bin/centroid_map
```diff
@@ -126,14 +133,14 @@ def centroid_map(mapfile, img, outfile):
         if 0 == uu:
             continue
         idx = np.where(mapvals == uu)
-        w = np.sum(img.imgvals[idx])
+        w = np.nansum(img.imgvals[idx])
         if 0 == w:
             # If sum is 0, use unweighted value
             cx = np.average(img.xx[idx])
```
DougBurke, line 139: do we not need to worry about nan's in this case? eg use `np.nanmean` rather than `np.average`?

DougBurke: Aha. I hadn't noticed the `img.imgvals` vs `img.xx` difference.

PR #860, opened 2024-03-01
*search_csc: initial csc2.1 support/branding

kglotfelty:


DougBurke:
I'm hoping we should get more clarity next week on the new endpoints (but I'm not sure that's relevant here, other than checking the tag name).

kglotfelty:
I also don't know if the list of filetypes is actually correct for 2.1 (for current I just set it to use the 2.0; but didn't check if that was right).  Similarly I don't know if the pre-canned/"default" search results are the same as 2.0 -- will need to see what cscview comes out with when it's actually released.

Probably should have put that in the initial PR comment.

DougBurke:
So, we now have apec results, so there may be a tweak to the spectral data. Other than that I'd expect the queries to be similar (but there's also tweaks to variability, so maybe that will change the variability query). Hmmm.

kglotfelty:
and I'm assuming still no CLI i/f to limsens.

DougBurke:
It looks like we already have CSCCLI access but just **not** advertised ...

```
% wget -O limsens.vot 'http://cda.cfa.harvard.edu/csclimsen/limsen?RA=246.7925&DEC=-24.32027&SR=0.05'
```

where `SR` is in degrees.

kglotfelty:
Last I heard ... that's not supposed to be "public" ... so not something we should be using

See https://icxc.harvard.edu/pipe/catdev/2021/0520.html

kglotfelty:
Humm... and I presume that what's there will need some kind of version/release parameter added?  (Or separate 21 endpoint?)

DougBurke:
That is a question I have yet to had people understand why I ask it...

DougBurke:
I'm going to take this rather than #865 since I think the decision to enable this is likely to have to be done post the release, what with all the other stuff going on.

PR #870, opened 2024-04-03
removed an extraneous whitespace

nplee:
take it or leave it for 4.16.1; just noticed extra whitespace preceding a `==`

### DIFF
--- a/bin/specextract
+++ b/bin/specextract
@@ -2478,7 +2478,7 @@ cannot contain any spaces")
 
 
 
-if __name__  == "__main__":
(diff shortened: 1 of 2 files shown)


DougBurke:
I don't think we want to take this for 4.16.1 (it doesn't provide any benefit) but we should take once that has been released.

nplee:
added new commit to fix a 'zero-counts in region' check that was missed (from helpdesk #025335) since a `NoneType` variable was unintentionally being evaluated as a boolean and skipping over the test in the conditional.

PR #874, opened 2024-04-17
cgs: fix problem with PHAII files that do not contain background spectra

kglotfelty:
This fixes problem reported in CXC hd ticket 25247.

The tgcat PHA2 files don't always have background spectra.  The script handles this case mostly ok, but is failing when it tries to .sort() a list with all `None`s.   Not sure if this a change in python behavior, but trivial to fix.  (Only sort if values are not `None`).

### DIFF
--- a/bin/combine_grating_spectra
+++ b/bin/combine_grating_spectra
@@ -178,15 +178,22 @@ def split_II_to_I( instk, arfs, rmfs, outroot, verbose, clobber ):
         outs = [delme(x) for x in glob.glob( tmproot+"_?eg_??.pha")]
         bkgs = [delme(x) for x in glob.glob( tmproot+"_?eg_??_bkg.pha")]
 
+        outs.sort()
         retsrc.extend(outs)
 
         if ( 0 == len(bkgs) ):
             retbkg.extend([None]*len(outs))
-        else:
+        elif len(bkgs) == len(outs):
+            bkgs.sort()
             retbkg.extend(bkgs)
-            retbkg.sort()
+        else:  # should never happen but just in case
+            raise IOError("Mismatch in number of source and background PHA files")
 
-    retsrc.sort()
+    for sss,bbb in zip(retsrc,retbkg):
+        if bbb:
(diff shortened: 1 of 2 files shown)


bin/combine_grating_spectra
```diff
@@ -184,9 +184,9 @@ def split_II_to_I( instk, arfs, rmfs, outroot, verbose, clobber ):
             retbkg.extend([None]*len(outs))
         else:
             retbkg.extend(bkgs)
-    
+            retbkg.sort()
```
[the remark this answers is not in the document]
kglotfelty: Thanks.  Agree there is a potential for this to get messed up if user supplies a mix of type:II pha files w/ and w/o backgrounds.  

I've reworked the logic and added a sanity check.

bin/combine_grating_spectra
```diff
@@ -178,15 +178,22 @@ def split_II_to_I( instk, arfs, rmfs, outroot, verbose, clobber ):
...
-
-    retsrc.sort()
+        else:  # should never happen but just in case
+            raise IOError("Mismatch in number of source and background PHA files")
+
+    for sss,bbb in zip(retsrc,retbkg):
+        if bbb:
+            sroot = sss.rstrip(".pha")
+            broot = bbb.rstrip("_bkg.pha")
+            assert sroot == broot, "Something is messed up in sort/bkg matching"
```
DougBurke, line 196: If this were to happen I assume there's nothing a user could do (e.g. can they explicitly list the files instead; it's too early for me to trace through to know exactly where this is happening to know if any of this is potentially user-controlled)?

kglotfelty: Correct.  Basically that would be a bug in `tgsplit`.  It should either produce no background or as many background as source files ; if there's a diff then we have to go back `tgsplit` and figure out what's wrong.

DougBurke: So, in other words it's not worth cleaning up the error message.

kglotfelty: I don't this so.

PR #879, opened 2024-05-07
chandra_repro: tweak for pre-repro3 (very early mission) data

kglotfelty:
Vinay found an observation in the archive that has not been through repro3 so the *FILE keywords used to automatically locate auxillary files are missing. Ref hd tix 25280.

This PR updates the way `chandra_repro` calls `r4_header_update` to specify the pbk and asol files names.  This works whether the *FILE keywords are present or not.  (There is other logic in chandra_repro that already checks that the *FILE keywords matches what it finds when globbing through the primary and secondary dirs.)

### DIFF
--- a/bin/chandra_repro
+++ b/bin/chandra_repro
@@ -1,9 +1,8 @@
 #!/usr/bin/env python
 
 #
-# Copyright (C) 2010-2022  Smithsonian Astrophysical Observatory
-#
-#
+# Copyright (C) 2010-2022, 2024 
+# Smithsonian Astrophysical Observatory
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


DougBurke:
Ahh. I had assumed Vinay was just using an old dataset he had lying around!

kglotfelty:
Update:  OK, so **the** problem isn't missing *FILE keywords.  The problem is that the DM is fouling up the GTIs.  

r4_header_update uses the `ObsInfo` class to locate the files, but after running a_p_e, the GTIs are even more fouled up and it ends up creating a generic `GTI` blocks w/o a CCD_ID keyword which throws an exception in the ObsInfo __init__. 

The "real" fix will need to happen in the DM; however, this change, to specify the file names explicitly is a good work around, so still a useful PR.

PR #885, opened 2024-07-03
minor tweak to build XSpec user model on CIAO 4.17b1

nplee:
> It should be an easy fix, he says after joyfully forgetting all the previous times an XSPEC change should have been easy... (@DougBurke)

Fortunately, it was in this case.  I tested this tweak out with the 'relxill' v2.3 model and was able to get it to build on Conda CIAO 4.17b1 by omitting "XSModel" from the list of libraries expected by the script.

### DIFF
--- a/bin/convert_xspec_user_model
+++ b/bin/convert_xspec_user_model
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 #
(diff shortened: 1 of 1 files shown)


DougBurke:
So, we either start a 4.17 branch now or write this in a way that can be used with both 4.16 and 4.17 code. Hmmm.

nplee:
I've just pushed a change to check CIAO version to handle this and avoid needing a new branch.

PR #887, opened 2024-07-09
sherpa_contrib.diag_resp

nplee:
New helper script to simplify generating a diagonal RMF, and associated flat ARF, data objects in Sherpa by specifying various telescope/instrument configuration or using a reference spectrum.  By default, it generates responses for Chandra/ACIS PI-formatted spectra.

Should the naming be more specific (e.g. `diag_resp` > `diagonal_response`)?


sherpa_contrib/diag_resp.py
```diff
@@ -0,0 +1,712 @@
+#
+#  Copyright (C) 2024
+#           Smithsonian Astrophysical Observatory
+#
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
```
kglotfelty, line 8: version **3**

sherpa_contrib/diag_resp.py
```diff
@@ -0,0 +1,712 @@
...
+
+
+
+class EGrid:
+    """
+    gather together all information for non-Chandra setups and return the energy-grid.
+    """
+
+    def __init__(self, telescope, instrument, detector, instfilter, nchan, chantype):
+        self.telescope = telescope
```
kglotfelty, line 144: I'd suggest you go ahead and `.lower()` and `.upper()` the values once in the constructor instead of each time you go to use them

sherpa_contrib/diag_resp.py
```diff
@@ -0,0 +1,712 @@
...
+    revise telescope/instrument configuration information to the
+    minimal parameters required to obtain the necessary energy grid
+    """
+
+    if telescope == "" or telescope is None:
+        raise ValueError("'telescope' parameter must be specified!")
+
+
+    ### update telescopes with alternate names ###
+    if telescope.upper() == "XTE":
```
kglotfelty, line 320: Instead of these cascading `if` statements, you might want to use a dictionary instead

```python
valid_vals = {'xte': 'RXTE', 'sax' : 'BeppoSAX', 'erosita': 'SRG'}
if telescope in valid_vals:
  telescope = valid_vals[telescope]
else:
  raise ValueError(...stuff...)


```

sherpa_contrib/diag_resp.py
```diff
@@ -0,0 +1,712 @@
...
+    """
+    Return a diagonal RMF and flat ARF data objects with matching energy grid for a specified instrument/detector.
+    Use set_rmf and set_arf on the respective instances.  Use 'build_resp' for a non-instrument specific [or generalized] energy-grid configuration.
+
+    "PI" channel-type is the default, assumed spectral type.
+
+
+    Supported arguments:
+
+        telescope="Chandra"
```
kglotfelty, line 554: Gratings?

sherpa_contrib/diag_resp.py
nplee: addressed in commit bad862a80b9d2a79a7c5aa7718962c30d28dc5af

sherpa_contrib/diag_resp.py
nplee: addressed in commit bad862a80b9d2a79a7c5aa7718962c30d28dc5af

sherpa_contrib/diag_resp.py
nplee: I need to think this through... what would be the use case (since it's still a bit unclear to me if there's a use with the stowed background alongside dispersed data set?  That said, ObsIDs 6647 and 6648 are ACIS-S/HETG observations that had the zeroth-order coordinates incorrectly input  by the PI, so much of the field is source free...)

Also, how many diffraction orders should we go out?  It's easy enough to get 0-3; but any higher?

kglotfelty: IDK;  I just noticed they weren't covered.

Presumably same use case as say the other non-chandra datasets :)

In general the grating RMFs are "nearly" diagonal -- but take a while to compute; so one may want a diagonal response to get a quick-n-dirty answer.

nplee: fair 'nuff! I think I've just settled on @DougBurke's school of thought: if you're using gratings you have too many counts to play with and it's your problem.

::continues gnawing::

sherpa_contrib/diag_resp.py
```diff
@@ -0,0 +1,908 @@
...
+
+        warnings.formatwarning = warnings_reset
+
+        return func_out
+
+    return run_func
+
+
+
+def _quash_shpverb(func):
```
DougBurke, line 94: You should be able to do this with just

```
from sherpa.utils.logging import SherpaVerbosity
...
with SherpaVerbosity("ERROR"):
    ... call code ...
```

So, either this can be used in the decorator, or you can avoid the decorator and use the form above (I haven't checked how it's used).

This is not something that needs to be done now, since the code works as is.

sherpa_contrib/diag_resp.py
```diff
@@ -0,0 +1,908 @@
...
+    from string import ascii_letters,digits
+    from random import choices
+
+    chars = ascii_letters + digits
+
+    return ''.join(choices(chars, k=strlen))
+
+
+
+def _arg_case(arg: str|int|None=None,
```
DougBurke, line 126: I don't think this will work with python 3.9, since the `|` form was, I believe, added in Python 3.10.

So change to `...(arg: Optional[Union[str, int]],` and `...) -> Optional[str]` and add a `from typing import Optional, Union` (assuming they are not already imported).

sherpa_contrib/diag_resp.py
```diff
@@ -0,0 +1,908 @@
...
+
+
+
+@_quash_shpverb
+def _get_file_header(fn) -> dict:
+    from sherpa.astro.ui import load_pha,get_data,delete_data
+
+    tmp_id = _get_random_string(strlen=32)
+
+    load_pha(tmp_id,fn)
```
DougBurke, line 144: We can avoid the need for a random string here since we should be able to just import `unpack_pha` instead of the other symbols and say

```python
return unpack_pha(fn).header
```

sherpa_contrib/diag_resp.py
DougBurke: There are other lines below which will need updating too.

sherpa_contrib/diag_resp.py
```diff
@@ -0,0 +1,908 @@
...
+                                           instrument = self.instrument,
+                                           chantype = self.chantype)
+
+        blkname = self._get_ebounds_blk(instrument = instrument,
+                                        detnam = detnam,
+                                        instfilter = instfilter,
+                                        subchan = subchan,
+                                        chantype = chantype)
+
+        try:
```
DougBurke, line 273: So, we should be able to just use the sherpa io code to read this in (in a round-about manner). The following has not been tested

```
from sherpa.astro.io import read_table_blocks
```

For example

```
sherpa-4.16.0> from sherpa.astro.io import read_table_blocks

sherpa-4.16.0> infile = "data/ebounds-lut/xrism-ebounds-lut.fits"

sherpa-4.16.0> _, blocks, hdrs = read_table_blocks(infile)
```

Now, blocks and hdrs are dictionaries with keys 1, ..., nblocks - e.g.

```
sherpa-4.16.0> blocks.keys()
dict_keys([1, 2, 3, 4, 5])
```

Unfortunately it's not easy to find the block, since crates eats up the HDUNAME/EXTNAME values, so (for Crates) we have

```
sherpa-4.16.0> from sherpa.astro import io

sherpa-4.16.0> io.backend.__name__
'sherpa.astro.io.crates_backend'

sherpa-4.16.0> hdrs[2]
{'TELESCOP': 'HITOMI',
 'INSTRUME': 'SXI',
 'DETNAM': 'PIXEL',
 'FILTER': '',
 'CHANTYPE': 'PI',
 'HDUCLASS': 'OGIP',
 'HDUCLAS1': 'RESPONSE',
 'HDUCLAS2': 'EBOUNDS',
 'HDUVERS': '1.2.0',
 'HDUVERS1': '1.1.0',
 'HDUVERS2': '1.2.0',
 'RMFVERSN': '1992a',
 'CCLS0001': 'CPF',
 'CCNM0001': 'EBOUNDS',
 'CDTP0001': 'DATA',
 'CVSD0001': '2016-03-20',
 'CVST0001': '14:40:33',
 'CDES0001': 'Energy boundaries of spectral channels',
 'CISTATUS': '1',
 'DETCHANS': '4096',
 'CHECKSUM': 'Z5AYb42YZ49Yb49Y',
 'DATASUM': '3744066889',
 'EUNIT': 'keV'}
```

whereas if we used AstroPy we would get

```
sherpa-4.16.0> import sherpa.astro.io.pyfits_backend

sherpa-4.16.0> io.backend = sherpa.astro.io.pyfits_backend

sherpa-4.16.0> _, blocks2, hdrs2 = read_table_blocks(infile)

sherpa-4.16.0> hdrs2[2]["EXTNAME"]
'XTEND'

sherpa-4.16.0> hdrs2[2]["HDUNAME"]
'XTEND'
```

So, instead of creating `blkname` you'd have to do something like

```
found = None
for idx, hdr in hdrs.items():
    if hdr["INSTRUME"] != instrument or hdr["DETNAM"] != detnam or ...:
        continue
    found = idx
    break

if found is None:
    raise SomeSortOfError

coldata = blocks[idx]
hdrdata = hdrs[idx]
```

sherpa_contrib/diag_resp.py
```diff
@@ -0,0 +1,908 @@
...
+            instrument,detnam = runchecks(instrument,detnam)
+
+        return instrument, detnam
+
+
+    def _check_chandra(self,inst,det):
+
+        ### check for Chandra/HRC, which is barely suitable for crude hardness ratio estimates ###
+        if inst == "HRC":
+            raise RuntimeWarning("non-grating HRC data has insufficient spectral resolution to be suitable for spectral fitting!")
```
DougBurke, line 516: Don't let Vinay see this comment!

sherpa_contrib/diag_resp.py
nplee: having installed a 3.9 environment, you are unfortunately correct, and I have revised... it looked so nice using pipes

sherpa_contrib/diag_resp.py
nplee: I've been playing around `read_table_blocks` and the two IO backends, and I think it's a bit fragile if we're relying on the header keywords pulled from the original CalDB files... and they just don't necessarily provide sufficient information.  For example, the XMM file is absolutely bare except for the keyword I added to it when I extracted it:

```
falafel-119: dmlist xmm-ebounds-lut.fits keys
 
--------------------------------------------------------------------------------
Header keys for block EPIC-MOS_PI
--------------------------------------------------------------------------------
 
0001 EUNIT                eV                             String       energy grid units
```

and even the ACIS gratings file doesn't have a keyword to indicate the grating order...

```
falafel-122: dmlist chandra-ebounds-lut.fits keys
 
--------------------------------------------------------------------------------
Header keys for block ACIS-HEG/1
--------------------------------------------------------------------------------
 
0001 CREATOR              mkgrmf v0.2.4-1                String       Program creating this file
0002 DATE                 2024-07-15T17:08:21            String       Date and time of file creation
0003 FILTER                                              String       
0004 CHANTYPE             PI                             String       
0005 DETCHANS             8192                           Int4         
0006 HDUCLASS             OGIP                           String       
0007 HDUCLAS1             RESPONSE                       String       
0008 HDUCLAS2             EBOUNDS                        String       
0009 HDUVERS              1.3.0                          String       
0010 RMFVERSN             1992a                          String       Obsolete
0011 HDUVERS1             1.1.0                          String       Obsolete
0012 HDUVERS2             1.2.0                          String       Obsolete
0013 MISSION              AXAF                           String       Mission
0014 TELESCOP             CHANDRA                        String       Telescope
0015 INSTRUME             ACIS                           String       Instrument
0016 DETNAM               ACIS-7                         String       Detector
0017 OBS_ID               7280                           String       Observation id
0018 DS_IDENT             10.25574/07280                 String       Dataset Identifier: DOI
0019 READMODE             TIMED                          String       Read mode
0020 GRATING              HETG                           String       Grating
0021 CYCLE                P                              String       
0022 SIM_X                   -0.68282252473119           Real8        SIM focus pos
0023 SIM_Y                                   0           Real8        SIM orthogonal axis pos
0024 SIM_Z                     -187.1254020033           Real8        SIM translation stage pos
0025 CHECKSUM             A3TAD2R5A2RAA2R3               String       HDU checksum updated 2024-07-17T20:57:30
0026 DATASUM              718591150                      String       data unit checksum updated 2024-07-15T17:08:22
0027 EUNIT                keV                            String       energy grid units
```

So I'm not entirely convinced this is a more appropriate approach.  I went about handling pycrates vs. astropy separately because of the HDU name issues I was having, which you've highlighted here...

nplee: I might be a dolt, but maybe the easiest thing is to also write the block name as a new header keyword too when for these tables?  Then I'd be able to use `read_table_blocks`, agnostic of backend.

DougBurke: Yeah - the generic I/O routines are somewhat specific to Sherpa and not really designed as a general purpose library.

PR #888, opened 2024-07-11
dax: remove obsolete awk script

kglotfelty:
This awk script is no longer needed/used.  It was replaced w/ the `convert_ds9_region_to_ciao_stack` script.


DougBurke:
We should check to see if the conda install removes the old version (not really a problem if it doesn't, just useful to know what happens).

PR #889, opened 2024-07-11
MaskedIMAGECrate: add check for column ranges in addition to subspace/region

kglotfelty:
This update adds a check for the `range_min` and `range_max` on the axes.  This matches a recent-ish change to the C dmimgio routines that also now check the axes ranges.   Without it, there were cases where the top most and right most pixels at the image edge would be inside the subspace, but the center of the pixel would be outside the range. Ie a partial pixel. 

We need this range change to bring the python code in sync w/ the C code so they agree on which pixels are valid.

### DIFF
--- a/crates_contrib/masked_image_crate.py
+++ b/crates_contrib/masked_image_crate.py
@@ -72,13 +72,32 @@ class MaskedIMAGECrate(IMAGECrate):
             #  warning()
             return
 
-        try:
-            xform = self.get_axis(sky).get_transform()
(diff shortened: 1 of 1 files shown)


kglotfelty:
Humm ... this didn't merge correctly.  It only got the 2nd commit e95c09 .  The first commit 387f9a9 is missing.

I'll make a new PR to try to pick up the change.

kglotfelty:
nevermind; driver error.

PR #893, opened 2024-07-16
merge_obs & friends: drop use of dmcopy opt="all" due to GTI bug

### LINKED ISSUE #758: check for dmcopy opt=all
Looking at the `merge_obs` helpdesk ticket re: failing due to pixel size diffs , I noticed that it is using `dmcopy` with `opt=all` .  We should review if that's really needed/necessary.  It can definitely be problematic due to the GTI/subspace bug if the user has applied time filters.

For gratings data ... since we're reprojecting the **events** , the `REGION` block will not be valid anymore so carrying it around doesn't help users.

I also see we're using `dmcopy` with `opt=all` in the `fileio.find_output_grid` routine ... I don't think we need it there gratings or non-gratings.

kglotfelty:
This closes #758

### DIFF
--- a/ciao_contrib/_tools/merging.py
+++ b/ciao_contrib/_tools/merging.py
@@ -516,7 +516,7 @@ def _copy_region_block(infile, outfile):
     in_ds = inblk.get_dataset()
     try:
         regblk = in_ds.get_crate("REGION")
-    except IndexError as no_region:
+    except IndexError:
         return
 
     dmappend(f"{infile}[REGION][subspace -time]", outfile)
--- a/ciao_contrib/_tools/merging.py
+++ b/ciao_contrib/_tools/merging.py
@@ -507,6 +507,21 @@ def update_ancillary_keywords(infile, outfile, asolfile):
 #            # do this.
 #            # update_ancillary_keywords(infile, outfile, asolfile)
 
+def _copy_region_block(infile, outfile):
+    """DM bug with dmcopy opt=all means we need to avoid it and
+    copy any REGION blocks manually (gratings data)"""
+    from ciao_contrib.runtool import dmappend
+
+    inblk = pycrates.read_file(infile)
+    in_ds = inblk.get_dataset()
+    try:
+        regblk = in_ds.get_crate("REGION")
+    except IndexError as no_region:
+        return
+
+    dmappend(f"{infile}[REGION][subspace -time]", outfile)
+
+
 def reproject_events_task(infile, ra0, dec0, outfile,
                           tmpdir="/tmp/",
                           verbose=0,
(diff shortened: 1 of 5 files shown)


ciao_contrib/_tools/merging.py
```diff
@@ -507,6 +507,21 @@ def update_ancillary_keywords(infile, outfile, asolfile):
...
+def _copy_region_block(infile, outfile):
+    """DM bug with dmcopy opt=all means we need to avoid it and
+    copy any REGION blocks manually (gratings data)"""
+    from ciao_contrib.runtool import dmappend
+
+    inblk = pycrates.read_file(infile)
+    in_ds = inblk.get_dataset()
+    try:
+        regblk = in_ds.get_crate("REGION")
+    except IndexError as no_region:
```
DougBurke, line 519: Since we are ignoring the exception we can just say

```suggestion
    except IndexError:
```

ciao_contrib/runtool.py
```diff
@@ -3468,7 +3468,7 @@ def make_tool(toolname):
...
+    'opt': [ParValue("columns","s","List of columns to include",'INDEF'),ParSet("download","s","Download data products for which sources?",'none',["none","ask","all"]),ParValue("root","f","Output root for data products",'./'),ParValue("bands","s","Comma separated list of CSC band names taken from broad, soft, medium, hard, ultrasoft, wide. Blank retrieves all",'broad,wide'),ParValue("filetypes","s","Comma separated list of CSC filetypes.  Blank retrieves all",'regevt,pha,arf,rmf,lc,psf,regexp'),ParSet("catalog","s","Version of catalog",'csc2.1',["csc2.1","csc2","csc1","current"]),ParRange("verbose","i","Tool chatter level",1,0,5),ParValue("clobber","b","Remove existing outfile if it exists?",False)],
```
DougBurke, line 3471: The changes to `obsid_search_csc` and `search_csc` shouldn't really be here (I know we are going to regenerate the files anyway, so it's not really a terrible change).

DougBurke:
I have to admit to not remembering why I added these (or some of these) in the first place, but I am pretty sure that some of them were needed, back in the day. I am pretty sure that it then became a defensive posture (e.g. "better do this just in case"), so some cargo-culting was likely involved.

kglotfelty:
the `find_output_grid`definitely isn't needed since it's just a temp file used to seed `skyfov`.

The other two places will copy the `REGION` extension via `dmappend`.  For the `reproject_obs` and `merge_obs` thread this really shouldn't happen because the physical WCS in the `REGION` extension will no longer match that in the event block; I'm not 100% sure this change hits that particular use case but I left it in there just in case.  [Note: reproject_events doesn't copy all blocks so there's a bit of a disconnect between the no-reproject and w/ reproject paths in the code.]

PR #897, opened 2024-08-19
Clearify aplimits examples

hamogu:
This is in response to a helpdesk user question asking when the limits returned are per exposure vs. per second.


DougBurke:
Ta

PR #899, opened 2024-08-20
Practical advice on how to use aplimits

hamogu:
This is a follow-on to #897. The helpdesk ticket that triggered that other PR came back with more questions, so I'm adding more text.

### DIFF
--- a/share/doc/xml/aplimits.xml
+++ b/share/doc/xml/aplimits.xml
@@ -56,6 +56,21 @@
         both bkg_rate and m are given, the tool assumes that the background rate
         is known and ignores the setting for m.
       </PARA>
+      <PARA>
+      In practice, one never has perfect background knowledge, but in many cases
+      the background rate can be determined well enough that the uncertainty does
+      not matter in practice. For an isolated point source, the background can be
+      measured from a large source-free region on the same chip. If the background
+      region is large enough to yield a large number of counts (50 or more), then
+      one can typically use background rate (number of counts in the
+      background region divided by the area and exposure time of the background region)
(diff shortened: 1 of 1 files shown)


share/doc/xml/aplimits.xml
```diff
@@ -56,6 +56,21 @@
...
+      measured from a large source-free region on the same chip. If the background
+      region is large enough to yield a large number of counts (50 or more), then
+      one can typically use background rate (number of counts in the
+      background region divided by the area and exposure time of the background region)
+      as known background rate. For extended sources, crowded fields, or exposures in
+      sub-array mode, it might not be possible to find a large enough background
+      region. In that case, the user should find the best possible uncontaminated
+      background region and use the number of counts in that region as input for m
+      (and also set the the geometric areas (A_s, A_b) and respective exposure times
+      (T_s, T_b) is needed).
```
DougBurke, line 72:
```suggestion
      background region and use the number of counts in that region as input for m,
      and also set the the geometric areas (A_s, A_b) and respective exposure times
      (T_s, T_b).
```

This seems a small-bit better to me, but have I lost the point with this change? If so then we'd want to change `is needed` to `as needed`, I guess.

hamogu, line 72: How about that (which was my original intention):
```suggestion
      (T_s, T_b) if needed).
```

But know I wonder if the user would ask: How do I know if it's needed?
So I think I'll just go with your suggestion.

DougBurke: That was my concern with a comment like "if needed" !

PR #901, opened 2024-08-22
mkrprm: make radial profile response matrix

kglotfelty:
New script to create the equivalent of a response matrix for radial profile; that is the fraction of the PSF that is included in adjacent radial bins.  Based on notebook/discussion with @hamogu a while ago (like several years).

The output is a 2D array, ie image, ie response matrix.

Hopefully can replace/extent-to using mkpsfmap+dmimgadpt w/ marx when offset issue is fixed.

There is also a new sherpa convolution model: `MatrixModel` which will multiply (ie conovolve) with a matrix.

```python
import sherpa.astro.ui as ui
import numpy as np

# Create an identity matrix
from sherpa_contrib.matrix_model import MatrixModel

from pycrates import read_file
# ui.load_data(1, "radial.prof[cols rmid,sur_bri,sur_bri_err]", Data1D)
ui.load_data(1,"radial_profile.fits",3,["RMID","SUR_FLUX","SUR_FLUX_ERR"])

response_matrix = read_file("frac.out").get_image().values*1.0
x_vals = ui.get_data().x
mymatrix = MatrixModel(response_matrix, x_vals, name="rprm")

pp = ui.polynom1d("poly")
ui.thaw(pp.c1)
pp.c1 = -1.0
pp.c0 = 12.5

ui.set_source(mymatrix(pp) )
ui.fit()
ui.plot_fit_delchi()

```

The `MatrixModel` is definitely rough and there are some oddities that will need to be worked out.


kglotfelty:
# Problem with `MatrixModel` sherpa model when using notice/ignore

The `MatrixModel` needs to evaluate the model over the entire _dataspace_ and the data grid must match the response matrix grid. When you `notice` or `ignore` a subset of bins, it breaks.

```python
import numpy as np
import sherpa.astro.ui as ui
from matrix_model import MatrixModel

npts = 10
xx = np.arange(npts)+10
yy = np.random.rand(npts)
ee = np.ones_like(xx)*0.01

ui.load_arrays(1, xx, yy, ee, ui.Data1D)

m_model = MatrixModel(np.identity(npts))

ui.set_model(m_model(ui.const1d.c0))
ui.fit()

ui.notice(12,17)
ui.fit()
```
The first `fit()` works, the 2nd `fit()` fails.

Not sure how to get the full dataspace grid?

I don't think a show stopper for release but would be nice to address.

PR #902, opened 2024-09-03
update packaging and installation functionality

nplee:
- updated `mk_script_tarfile` to default to Python 3.11 for CIAO 4.16 (or use the Python version being used with the CIAO that's active while building the tarball)
- `setup.py` updated to include `data/ebounds-lut` so that the package can be installed with `pip install` from a local repository (brought to attention by @kglotfelty).

### DIFF
--- a/mk_script_tarfile
+++ b/mk_script_tarfile
@@ -34,7 +34,7 @@ import sys
 import time
 
 
-def build_dist(version, dirname, outdir, python="python3.10"):
+def build_dist(version, dirname, outdir, python="python3.11"):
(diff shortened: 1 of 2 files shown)


DougBurke:
I think we should be able to go to a more-declarative system and use `pyproject.toml` (possibly even moving away from `setuptools`), but that's for another day...

nplee:
quick comment, while I was looking into this, I also experimented with the exclusive `pyproject.toml` approach and if we continue using `setuptools`, we'll need to jump the base version we use significantly at the very least.

PR #911, opened 2024-10-16
Update xcm

DougBurke:
There are changes in Sherpa 4.17.0 which break the xcm.notice/ignore call. We can actually fix this before 4.17.0 is released, and also improve the internal mapping of XSPEC group number to channel range (XSPEC appears to use the GROUPING and QUALITY values when grouping whereas Sherpa only uses GROUPING).

One result is that we now get more warnings when using a PHA file where the QUALITY value changes within a GROUP, but it is technically more correct now.


DougBurke:
@kglotfelty - this will hopefully fix the SDS regression tests failures in CIAOX.

PR #916, opened 2024-10-28
Diagonal response for Sherpa 4.17 & Python 3.10+

nplee:
- typo on 'einsten' fixed in tool and associated LUT file name fixed.
- identical bin edges of Einstein/SSS leads to immediate failure when used with create_arf/create_rmf; added function to replace the zero-valued edges to incrementally increase monotonically in steps of `dE~1e-16` above the `ethresh` value.
- use Python 3.10+ function typing formatting.
- pytests are updated to pass rather than xfail.


sherpa_contrib/diag_resp.py
```diff
@@ -637,13 +638,11 @@ def build_resp(emin, emax, offset: int, ethresh: Optional[Union[float,None]] = 1
...
-        # major,minor,micro = shpver.split(".")
-        # ver = float(f"{major}.{minor}")
-        # ver_micro = int(micro)
-        #
+        major,minor,micro = shpver.split(".")
+        ver = float(f"{major}.{minor}")
+        #ver_micro = int(micro)
 
-        if offset != 1:
+        if ver < 4.17 and offset != 1:
```
DougBurke, line 669: Technically this should not be necessary (as this should only ever be used with 4.17) but not worth changing.

Also, you can take advantage of tuple ordering rather than convert to a float - e.g

```
>>> ver = (4, 17)
>>> if ver < (4, 17):
...     print("Boo yah")
... 
>>> if ver <= (4, 17):
...     print("Boo yah")
... 
Boo yah
```

Again this doesn't need to be changed here.

DougBurke:
Is it worth adding anything to the changelog and/or ahelp file here (it would have to be in a separate PR)? No hurry if we do.

nplee:
I don't think so, and we can do that when we're prepping for the contrib 4.17.0 release, because honestly, who's really running the pytests besides us or using this unadvertised functionality with Einstein (or the new Einstein probe) data at this time?

PR #917, opened 2024-10-29
distutils/setuptools 'msvccompiler' dependency

nplee:
From a recent helpdesk ticket, `convert_xspec_user_model` uses `pip` to compile and install the model.  `pip` is ultimately calling on `numpy.distutils` modules that points to `numpy.distutils.mingw32ccompiler` which has a dependency on `distutils.msvccompiler`.

In `setuptools` 74.0.0 (https://setuptools.pypa.io/en/latest/history.html#v74-0-0), which reflects `distutils` development, it drops this functionality which subsequently leads to `pip` failing to build the model.  Adding an additional restriction of `setuptools < 74` in `convert_xspec_user_model` works around the issue.

This affects both CIAO 4.16 and 4.17 beta(s).

### DIFF
--- a/bin/convert_xspec_user_model
+++ b/bin/convert_xspec_user_model
@@ -67,7 +67,7 @@ for a description of the model.dat format.
 """
 
 toolname = "convert_xspec_user_model"
-toolver  = "4 July 2024"
+toolver  = "29 October 2024"
 
 import sys
 import os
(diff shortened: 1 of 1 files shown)


DougBurke:
So, I just don't understand why it ever looks for this. I would rather not add a n upper limit here, as it indicates a long-term issue with this tool, but it may be the best we can do just now.

It does remind me that we can probably replace the `oldest-suported-numpy` "hack" now though.

DougBurke:
I think the better idea wold be to stop `bin/convert_xspec_user_model` from importing `distutils` modiules (both top-level and `numpy`) but that requires some work:

https://github.com/cxcsds/ciao-contrib/blob/c600fdb1fa8cc62d22a51352d31f9230a69b3847/bin/convert_xspec_user_model#L775-L789

DougBurke:
I'm going to claim this is resolved by #923

PR #918, opened 2024-10-31
fall back on 'spawn' if 'fork' method fails in 'parallel_pool_futures'

nplee:
This update to the `parallel_pool_futures` function to fall back on the `spawn` multiprocessing context method if the default `fork` method fails to run, which is most likely to happen on MacOS, although fortunately hasn't been seen to happen frequently with CIAO, and this update is in response to the first such instance reported through the helpdesk.  While `spawn` is slower than `fork`, for the cases where `parallel_pool_futures` is [or can be] used in the CIAO scripts, the additional overhead to set up the parallelization with `spawn` will still be quicker than running the tasks serially (e.g. calculating the weighted source and background ARFs and RMFs simultaneously rather one-after-the-other).  

This update also linted lingering whitespace in the code.

### DIFF
--- a/ciao_contrib/parallel_wrapper.py
+++ b/ciao_contrib/parallel_wrapper.py
@@ -1,11 +1,11 @@
 #
-#  Copyright (C) 2011, 2015, 2016, 2019-2022
+#  Copyright (C) 2011, 2015, 2016, 2019-2024
 #                Smithsonian Astrophysical Observatory
 #
 #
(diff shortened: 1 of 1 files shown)


DougBurke:
We're going to need some sort of a README entry I think.

DougBurke:
Also I should talk to you to go through this change as I don't understand it. ..

PR #919, opened 2024-11-07
xcm.notice - better handle 4.17.0 filtering

DougBurke:
Thanks to changes to better support CHANNEL values not starting at 1 in 4.17.0 we need to rework the notice/ignore handling (which in XSPEC uses group numbers and not channels).

I had been converting manually to energy or wavelength filters, but it actually makes sense to keep the filtering in channel units. There is some work needed to try and replicate the UI fitering behaviour (that reports the filter change).

This is a follow-on to the failed attempt in #911


DougBurke:
@nplee - I need to think about this some more, which will have to wait until next week now. It hopefully fixes the issues you saw in the `convert_xspec_script` checks (but there's enough moving parts here that it may avoid the crash but still create subtly-different output).

PR #925, opened 2024-11-25
download_obsid_caldb: get p2_resp files for all temperatures

### LINKED ISSUE #920: download_obsid_caldb update for temperature dependent p2_resp files
Need to figure out logic being used with multiple p2_resp files 

https://github.com/cxcsds/ciao-contrib/blob/3441cc58e5a87a7640da1b48b0fb60f5f89343ef/bin/download_obsid_caldb#L466

The script will just retrieve all p2_resp files matching the date rather than say taking in MTL file and figuring out which FP_TEMP values to cherry-pick.

kglotfelty:
This fixes #920 .

Still needs ahelp update.

Confirmed works w/ current caldb and with Dale's test_3 w/ fp_temp dependent files.

Wish I could figure out a way to do it w/o needing to have range hard-coded, but doesn't look like caldb module will let me.

### DIFF
--- a/bin/download_obsid_caldb
+++ b/bin/download_obsid_caldb
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 #
-# Copyright (C) 2013-2022 Smithsonian Astrophysical Observatory
+# Copyright (C) 2013-2022, 2024 Smithsonian Astrophysical Observatory
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -25,7 +25,7 @@ import ciao_contrib.logger_wrapper as lw
 
 
 toolname = "download_obsid_caldb"
-__revision__ = "25 February 2022"
+__revision__ = "25 November 2024"
 
 lw.initialize_logger(toolname)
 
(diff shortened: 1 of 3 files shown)


bin/download_obsid_caldb
```diff
@@ -463,7 +462,8 @@ def get_tools(infile):
                ### 'rmf_file' : None,  # ???
                ### 'gain_file' : None # ???
                }
-    mkacisrmf = {'sc_matrix' : None, }  # infile
+    temps = ",".join([str(x) for x in range(153,178,2)])
```
DougBurke, line 465: I assume your complaint is that there's no way to automate creating the list of temperatures to use here, soit has to be hard-coded?

kglotfelty: I had hoped I could set `fp_temp=None` or _blank_ or `""` or even `delattr` it to remove any constraint; but none of those worked.

Honestly I'm shocked that a comma separated list works (or as it turns colons work too, who knew?).  Just a bit of maintenance PITA if the range changes and/or the granularity (step size) changes.

share/doc/xml/download_obsid_caldb.xml
```diff
@@ -355,6 +355,16 @@ they are skipped and only the background files are actually retrieved.
 
     </PARAMLIST>
 
+    <ADESC title="Changes in the script 4.17.0 (December 2022) release">
```
DougBurke, line 358:
```suggestion
    <ADESC title="Changes in the script 4.17.0 (December 2024) release">
```

PR #932, opened 2024-12-01
fine_astro: look for .gz aux files when copying

kglotfelty:
The `get_ancillary` method returns the path+filename w/o the `.gz` extension. 

```python
from ciao_contrib._tools import merging
import stk as stack

obsid = "26732"

files_in_dir = stack.build(f"{obsid}/primary/*,{obsid}/secondary/*")
print(files_in_dir)

obi = merging.validate_obsinfo(obsid)
aux_file = obi[0].get_ancillary("mask")
print(aux_file)

if aux_file not in files_in_dir:
    print("File not found, trying .gz")    
    assert aux_file+".gz" in files_in_dir, "Totally missing"
    print("Compressed version of file found.")
```

The DM does some magic to find `.gz` files if they exist, but we're using system copy so we need to check for missing extension before copying.

### DIFF
--- a/bin/fine_astro
+++ b/bin/fine_astro
@@ -305,6 +305,11 @@ def copy_auxfiles(obsid, out_dirs):
 
     for aux in aux_list:
         aux_file = obsid.get_ancillary(aux)
+        if not os.path.exists(aux_file):
+            aux_file += ".gz"
+            if not os.path.exists(aux_file):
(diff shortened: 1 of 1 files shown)


DougBurke:
We could move this logic into `ciao_contrib._tools.obsinfo` but that would need some thinking about, so it can happen "laterz™".

PR #941, opened 2025-01-30
download_obsid_caldb: fix p2_resp look by adding a range of FIDELITY …

kglotfelty:
…values

### DIFF
--- a/share/doc/xml/download_obsid_caldb.xml
+++ b/share/doc/xml/download_obsid_caldb.xml
@@ -357,8 +357,8 @@ they are skipped and only the background files are actually retrieved.
 
     <ADESC title="Changes in the script 4.17.1 (February 2025) release">
       <PARA>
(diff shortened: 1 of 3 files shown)


share/doc/xml/download_obsid_caldb.xml
```diff
@@ -355,6 +355,13 @@ they are skipped and only the background files are actually retrieved.
 
     </PARAMLIST>
 
+    <ADESC title="Changes in the script 4.17.1 (February 2025) release">
+      <PARA>
+        Fix to temperature dependent P2_RESP CALDB lookup due to a 
+        last minute change to how files are indexed.
```
DougBurke, line 361:
```suggestion
        Updates to fix retrieval problem for new temperature dependent
        P2_RESP files used by mkacisrmf.
```

I don't think users would gain much extra information with the original text...

kglotfelty:
confirmed working w/ 4.12 and  that 4.12 has been deployed to CDA server

PR #942, opened 2025-01-30
Fix issue 883

### LINKED ISSUE #883: find_chandra_obsid behaviour when there's no coverage
Why am I seeing this error message:

```
% find_chandra_obsid "a2472" v=4
Running: find_chandra_obsid
  version: 06 January 2023
Querying name resolvers for name=a2472
Found ra=340.47375 dec=17.54166667
    = 22h 41m 53.69999999999294s 17d 32' 30.000012000005256"
Querying Chandra archive for ra=340.47375 dec=17.54166667 radius=0.016666666666666666 inst=all grat=all
Retrieving URL: https://cxcfps.cfa.harvard.edu/cgi-bin/cda/footprint/get_vo_table.pl?strict=1&POS=340.47375,17.54166667&SIZE=0.016666666666666666 timeout=None
Search returned 0 rows
Chandra search found 0 results
After removing duplicates, have 0 matches
# find_chandra_obsid (06 January 2023): ERROR max() arg is an empty sequence
```

It should error out better than this. What has changed?

DougBurke:
I assume that something was changed in the 4.15.x series to deal with non-ASCII PI names which had unforseen consequences for the error checks. Or something has changed in how a null result is returned from the CXC Footprint Service. Otherwise I don't see how I could have not seen this problem.

Fix #883

### DIFF
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -1,4 +1,4 @@
-## 4.17.1 - Spring 2024
+## 4.17.1 - March 2025
 
 Updated scripts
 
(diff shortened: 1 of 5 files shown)


DougBurke:
Fixed conflict in changes file.

PR #943, opened 2025-02-04
Updated specextract warm_temperature_warning() to be consistent with …

dprincipe2:
…ACIS calibration and CIAO 417 documentation.  The trigger date has changed to 2018-01-01 and the ACIS focal plane temperature trigger has changed from -109 C to -111 C.

This change should be included in the next chance we have to make a new ciaocontrib package update.


nplee (approved):
changes look good; makes sense and are informational.

PR #945, opened 2025-02-07
Merge obs psfmap tmpstk

### LINKED ISSUE #935: merge_obs use stack for PSF merge infile
I hit an error when combining 45 observations because the `infile` parameter to `dmimgfilt` is longer than 2048 characters

```
  Combining 45 observations.
  Input string is longer than 2048 char: parameter argument error : infile
  # merge_obs (05 November 2021): ERROR Unable to run dmimgfilt with arguments: infile=...2070 characters....
```
I don't know if the error message is new or just first time hitting it.  Looking at the code, I think

https://github.com/cxcsds/ciao-contrib/blob/be347ee27ec259d848e8f682f88a37d40cb5ba1f/ciao_contrib/_tools/merging.py#L2873

can be updated to replace the long `infile` with a temporary stack and then it should be able to handle it.

Kenny

kglotfelty:
This fixes #935 by wrapping `dmimgfilt` call in new `TemporaryStack` context. 

This could be two PRs but would be linked so easiest to just do in one.

### DIFF
--- a/ciao_contrib/_tools/merging.py
+++ b/ciao_contrib/_tools/merging.py
@@ -49,6 +49,7 @@ import ciao_contrib._tools.run as run
 import ciao_contrib._tools.utils as utils
 
 from ciao_contrib._tools.headers import HeaderMerge
+from ciao_contrib.stacklib import TemporaryStack
 
 __all__ = (
     "match_obsid",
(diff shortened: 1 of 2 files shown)


DougBurke:
I wonder if `TemporaryStack` could just call `make_stackfile` rather than repeat the logic, but they are slightly different and I'm not sure I have the energy to merge the changes into `make_stackfile`. That can be a future-us problem.

PR #949, opened 2025-02-20
SimpleCoordTransform: fix for integer inputs (cast to float)

### LINKED ISSUE #947: SimpleCoordTransform cast input coordinates to float/reals
There is a bug (not a feature!) of the Transform module that if the input values are integers, then the output coordinates are also integers. 

The crates_contrib.utils.SimpleCoordTransform should force the input values to be float/real's to avoid this bug.  [And hopefully save someone else at least as many hours as it took me to realize why things were so messed up.]

kglotfelty:
This fixed #947

### DIFF
--- a/crates_contrib/utils.py
+++ b/crates_contrib/utils.py
@@ -898,7 +898,9 @@ class SimpleCoordTransform:
     def _convert(self, direction, mappings, x, y):
         """Convert a single (x,y) using the direction and mappings."""
 
-        pt = [[x, y]]
+        # Need to make sure x and y are float to avoid Transform bug
+        # when inputs are integer (forces output to be integer)
+        pt = [[x * 1.0, y * 1.0]]
         for trinfo in mappings:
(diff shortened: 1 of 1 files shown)


crates_contrib/utils.py
```diff
@@ -898,7 +898,9 @@ def _validate_coordsys(self, cname):
     def _convert(self, direction, mappings, x, y):
         """Convert a single (x,y) using the direction and mappings."""
 
-        pt = [[x, y]]
+        # Need to make sure x and y are float to avoid Transform bug
+        # when inputs are integer (forces output to be integer)
+        pt = [[x * 1.0, y * 1.0]]
```
DougBurke, line 903: I could argue that "send in an int, get an int back" makes sense, but probably not in Python...

kglotfelty: Not when going from logical to celestial.

PR #956, opened 2025-04-01
build XSpec user models for standalone Sherpa

nplee:
This is a minor modification to add support to the script to allow its usage with a standalone Sherpa build using a `--standalone` flag.

Since there's a single `ciao_contrib` dependency in the script for the logger wrapper, a user could just do a `pip install` from this repository to obtain the script and dependency and just `pip uninstall ciao-contrib` after the fact.

This update will makes the assumption that standalone Sherpa is built with XSpec model support and that the `HEADAS` variable is set in the environment.  From testing with the `relxill`, `reltrans`, and `warmabs` model packages, while building them with the Conda compilers is possible it was more straightforward to use the same compilers used to build XSpec that the models may have library dependencies on.

### DIFF
--- a/bin/convert_xspec_user_model
+++ b/bin/convert_xspec_user_model
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 #
-# Copyright (C) 2012 - 2016, 2020 - 2024
+# Copyright (C) 2012 - 2016, 2020 - 2025
 # Smithsonian Astrophysical Observatory
 #
 #
@@ -66,7 +66,7 @@ for a description of the model.dat format.
 """
 
 toolname = "convert_xspec_user_model"
-toolver  = "25 November 2024"
+toolver  = "31 March 2025"
 
 import sys
 import os
@@ -182,10 +182,15 @@ CXX_HEADER = """//
 """
 
 
-# The assumption is that by this point we know ASCDS_INSTALL must exist.
+# # The assumption is that by this point we know ASCDS_INSTALL must exist.
+# #
+# ASCDS_INSTALL = os.getenv("ASCDS_INSTALL")
+# if ASCDS_INSTALL is None:
+#     # _ = os.getenv("HEADAS")
+#     # ASCDS_INSTALL = f"{_}/../Xspec/{_.strip('/').split('/')[-1]}"
+#     ASCDS_INSTALL = os.getenv("HEADAS")
 #
-ASCDS_INSTALL = os.getenv("ASCDS_INSTALL")
-ASCDS_INSTALL_PATH = Path(ASCDS_INSTALL)
+# ASCDS_INSTALL_PATH = Path(ASCDS_INSTALL)
 
 
 def check_clobber(fname: Path) -> None:
@@ -943,7 +948,8 @@ def build_module_init(modname: str,
                       infiles,
                       extrafiles,
                       version,
-                      clobber: bool = False) -> None:
+                      clobber: bool = False,
+                      standalone: bool = False) -> None:
     """Create the module initalization code.
 
     Parameters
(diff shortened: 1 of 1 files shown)


bin/convert_xspec_user_model
```diff
@@ -1045,7 +1057,7 @@ __all__ = ({mdlnames})
                       f"  , 'buildinfo': {{ 'tool': '{toolname}', ",
                       f"'version': '{toolver}', ",
                       f"'sherpaversion': '{sherpa.__version__}', ",
-                      f"'ASCDS_INSTALL': '{ASCDS_INSTALL}', ",
+                      f"'{ascdsstr}': '{ASCDS_INSTALL}', ",
```
DougBurke, line 1060: This is setting up the metadata stored in the module to help track down problems. So we could just remove it. However, as is I do not see how this will work since `ASCDS_INSTALL` is no-longer set so shouldn't this raise a `NameError` here (also, I'm not sure what we want to store as the key for the standalone case).

DougBurke: Oh. I see `ASCDS_INSTALL` is set later in the `if __name__ == "__main__":` section. However, I am still not sure the key makes sense (or, rather, the names make this unclear). There's also the fact that `xspec-modelsonly` sets `HEADAS` in an "interesting" manner compared to a user who has manually built XSPEC, and I'd need to think about how that works here.

nplee: so I tested this using both a regular XSpec build and an `xspec-modelsonly` build and in both cases from what I can tell, the script seems to only be probing for the `libXS*` library files which are found in `$HEADAS/../Xspec/<build system info>/lib`.  From what I can tell, this is also how the `$ASCDS_INSTALL` location is being used... c.f.

```tcsh
falafel-107: ls $ASCDS_INSTALL/lib/libXS*
/soft/ciao-4.17/lib/libXSFunctions.so  /soft/ciao-4.17/lib/libXSUtil.so
/soft/ciao-4.17/lib/libXS.so

falafel-108: ls /scratch2/xspec-models/heasoft-6.35_xspec-12.15.0_models/Xspec/x86_64-pc-linux-gnu-libc2.28/lib/libXS*
/scratch2/xspec-models/heasoft-6.35_xspec-12.15.0_models/Xspec/x86_64-pc-linux-gnu-libc2.28/lib/libXSFunctions.so
/scratch2/xspec-models/heasoft-6.35_xspec-12.15.0_models/Xspec/x86_64-pc-linux-gnu-libc2.28/lib/libXS.so
/scratch2/xspec-models/heasoft-6.35_xspec-12.15.0_models/Xspec/x86_64-pc-linux-gnu-libc2.28/lib/libXSUtil.so

falafel-109: ls /scratch2/xspec/heasoft-6.35_xspec_12.15.0/Xspec/x86_64-pc-linux-gnu-libc2.28/lib/libXS*
/scratch2/xspec/heasoft-6.35_xspec_12.15.0/Xspec/x86_64-pc-linux-gnu-libc2.28/lib/libXSFit.so
/scratch2/xspec/heasoft-6.35_xspec_12.15.0/Xspec/x86_64-pc-linux-gnu-libc2.28/lib/libXSFunctions.so
/scratch2/xspec/heasoft-6.35_xspec_12.15.0/Xspec/x86_64-pc-linux-gnu-libc2.28/lib/libXSMinuit.so
/scratch2/xspec/heasoft-6.35_xspec_12.15.0/Xspec/x86_64-pc-linux-gnu-libc2.28/lib/libXSModel.so
/scratch2/xspec/heasoft-6.35_xspec_12.15.0/Xspec/x86_64-pc-linux-gnu-libc2.28/lib/libXSPlot.so
/scratch2/xspec/heasoft-6.35_xspec_12.15.0/Xspec/x86_64-pc-linux-gnu-libc2.28/lib/libXS.so
/scratch2/xspec/heasoft-6.35_xspec_12.15.0/Xspec/x86_64-pc-linux-gnu-libc2.28/lib/libXSStat.so
/scratch2/xspec/heasoft-6.35_xspec_12.15.0/Xspec/x86_64-pc-linux-gnu-libc2.28/lib/libXSUser.so
/scratch2/xspec/heasoft-6.35_xspec_12.15.0/Xspec/x86_64-pc-linux-gnu-libc2.28/lib/libXSUtil.so
```
which is tweaked in [L1629-L1647](https://github.com/nplee/ciao-contrib/commit/aaee928356cbc1e9cb2dbcb4a811fd17af990bef#diff-728c6f0d0f6302c4d129d71696d773c78f8f6b5240c157fdfc7e9e590bb5bc34R1629-R1647) in `mkname`

PR #959, opened 2025-04-14
Acis split evt by fptemp

kglotfelty:
@dprincipe2

I put together this script to automate the steps in the new P2_RESP/FP_TEMP thread.  When you get a chance can you take a look and see if this would be useful?

```bash
bin/acis_split_evt_by_fptemp
param/acis_split_evt_by_fptemp.par
```

would need a help file for release but hopefully self explanatory enough to check it out.

### DIFF
--- a/bin/acis_split_evt_by_fptemp
+++ b/bin/acis_split_evt_by_fptemp
@@ -22,7 +22,6 @@
 
 import os
 import sys
-import stk
 from pycrates import read_file
 from ciao_contrib.runtool import make_tool
 import ciao_contrib.logger_wrapper as lw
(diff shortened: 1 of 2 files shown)


DougBurke:
I was going for more the latter. Ta.

kglotfelty:
Sorry -- I didn't mean to blind side anyone. I wasn't really think this would make this upcoming script release. I was just working on the quarterly and after doing it once thought it was something that could possibly be automated. [And it's a lot more fun writing code than writing slides -- especially on a Friday afternoon :) ] 

Given how niche it really should be, I don't think it's a must have kind of thing; just something that could be useful for those corner cases where it could be beneficial.

dprincipe2:
@DougBurke @kglotfelty

I see now that the script doesn't actually apply specextract so its purpose is to just setup the files necessary for the user to run specextract. I think this is a useful tool for users. There are two caveats though:

(1) Some ACIS CCDs don't yet have any new p2_resp calibrations (e.g., most of ACIS-S) so users could run this expecting their data to receive some new calibration when it won't. That being said, the same thing could happen if users followed the CIAO thread without following the directions by looking at the included table which shows which CCDs are calibrated at which temps. Unfortunately, I don't see a way to easily automate this lookup in the script since we populate the warm p2_resp files with the -120C data if there are no 'new' warm calibrations for a particular CCD. I guess if you really wanted to mitigate this issue you could check the 'warm' p2_resp table data (specific columns) against the -120 C p2_resp table data and if the values are unchanged then it would demonstrate that particular ccd doesn't have new calibrations.  Ultimately though the resulting rmfs wouldn't be wrong, they would just not be what the user was expecting and the user could be unnecessarily slicing up their observation. The easiest way to avoid this issue without including the above check is to explain on whatever new thread or ahelp is included with the script release that only some CCDs benefit from this tool.  You could also reference the CIAO thread which the script is based off of which includes this info.

(2) I am planning to submit a spec for CIAO 4.18 to consolidate each CALDB p2_resp temperature bin file into one main p2_resp file and change mkacisrmf to generate responses at a user-specified temperature. If that eventually makes it into CIAO then this tool would have to be modified to account for a new single p2_resp file that includes all FP_TEMP calibrations

Given these two caveats, I'm in favor of releasing this now as long as we are ok with removing or modifying it down the line if/when the p2_resp file format gets changed.

Thanks @kglotfelty for putting this together!

kglotfelty:
Sorry to take so long to get back to this.

I added some checks to the beginning of the script to check to see if FI chips are in use.  I checked the CALDB files and don't see anything in the headers that specifies only 0-3 & 6 .  The script is currently set to check all the FI chips (0-4,6,8-9) .  I'm not a big fan of hard-coding specific calibration knowledge ... especially since it's unlikely users will be using 4, 8, or 9.  So I think this addresses the 1st concern. 

The 2nd thing I'll hold off on thinking about until we know more.  

Unless there are any objections I'll go ahead and mark ready to review and we'll get this into ciaox at some point.

dprincipe2:
Hi Kenny,

I understand your concern about hard-coding specific cal knowledge and I agree. If I understand though, your proposed check means that if the cal group does release new RMF temp calibrations for a BI chip (5,7), then the script won't allow that to proceed? I believe they are still working on calibrating the other chips for warmer temperatures but I'm not sure which ones. 

Would it be more prudent to allow the script to proceed for all chips but also print a warning with a link to our page or pages (https://cxc.cfa.harvard.edu/ciao/threads/acis-fptemp-resp/, https://cxc.cfa.harvard.edu/ciao/why/mkacisrmf.html) telling the users to check that their CCDs of interest are currently calibrated? Otherwise, we'd have to update the script if/when new cal files were released, right?

I don't feel strongly about this one way or another so feel free to go ahead with your method if you want. There is a real chance the cal group doesn't release any more files and, as you say, ccds 4, 8, and 9 don't get a ton of use in the last few years when temps have been more of a concern.  But unnecessarily slicing a dataset up does come with a cost  (temporal info) and it would be unfortunate if users thought they needed to do this for a source that ended up not having any new temp calibrations.

Thanks for tackling this issue!

kglotfelty:
Sorry I wasn't clear. They are warnings, not errors -- so if CAL releases fp+temp dependent BI cal files it'll still run w/ warnings. 

I can definitely link to the thread & why pages in the ahelp.

dprincipe2:
ah, ok. Sounds good then!

PR #971, opened 2025-06-12
support long region expressions in excess of 1023 characters

nplee:
This stems from helpdesk ticket #025802 where a user was running into a core dump.  I confirmed the behavior on both RHEL8 x86-64 and macOS 15 ARM64.  

There is a two-fold problem when the `srcreg` and/or `bkgreg` are used with region stacks in `srcflux`.  For sources near the chip edge `roi` (and subsequently the `splitroi` output used by `srcflux`) will return a source region intersected by a polygon region expression defining the detector FOV.  The polygonal region expression, let alone the compound expression, can easily exceed 1023 characters and when `srcflux` is used in this mode, it will first use `pycrates.read_file` to check whether or not the expression is a readable FITS file and then use `dmmakereg` to check the input region expression can be written in sky coordinates.

The core dump comes from `read_file`, which crashes when the input file expression has a long string length. Working around this, then `srcflux` then tries to use the region expression directly with `dmmakereg` fails due to a `region` parameter having a character length limit of 1023; putting the region expression in a temporary region file is then usable with `dmmakereg`.

A whole bunch of trailing white spaces are also eliminated.

### DIFF
--- a/bin/srcflux
+++ b/bin/srcflux
@@ -18,7 +18,7 @@
 #
 
 toolname = "srcflux"
-__revision__ = "17 April 2024"
+__revision__ = "10 June 2025"
 
 import os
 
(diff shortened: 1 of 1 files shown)


DougBurke:
Can the ahelp file be updated?

It would be nice if we didn't have to do this "guess whether things will fall over" dance, but I don't see how we can actually do this a better way (since we always are going to have some sort of a length limit). Oh well.

nplee:
ahelp file revised with change description.

PR #973, opened 2025-08-04
Dax tg coords arbitrary frame

kglotfelty:
interactive grating coords was only working if it was launched from the first, initial frame. Change now allows it to work in any frame.

Also fixes issue if prefs are set to CIAO region format, forces the region to be loaded in ds9 format.

### DIFF
--- a/bin/ds9_itgcoords_launch
+++ b/bin/ds9_itgcoords_launch
@@ -21,7 +21,7 @@
 
 import sys
 
-from dax.utils import xpaset, xpaget
+from dax.utils import xpaset, xpaget, xpaset_p
 
 
 def get_zero_order_location(xpa):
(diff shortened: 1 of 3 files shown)


DougBurke:
Is this ready to land?

kglotfelty:
Yes, it's in there in `chips_startup.tk`

```tcl
   169	
   170	    set tgcoords_wavelength [expr 12.39842 / $energy]
   171	
```

There was some wonkiness with which PR the tg_interactive coords got into ...I know I had something odd happen with the force-region-to-ds9-format change .  I'll double check more in the a.m. but first glance looks ok.

DougBurke:
So that line doesn't appear in the https://github.com/cxcsds/ciao-contrib/pull/973/files tab, but I have seen odd things with Sherpa PRs. As long as the text is okay then I'm happy.

Issue #954, opened 2025-03-08
What does the `chipx/chipy` output in `ciao_contrib/coords/chandra.py` mean?


hamogu:
Since Chandra dithers, the `chipx/chipy` coordinates of a source should depend on time. Yet, the `sky_to_chandra` does not take an asol file and a time as input and it's not clear how to encode that information in the `keyword_list`. For example, would I substitute the RA_NOM/RA_DEC keywords with the instantaneous values from the asol file? Or maybe RA_PNT/DEC_PNT? Or both? Additionally, it seems pretty cumbersome to do that to iterate the transformation over an entire photon list.

So, I wonder, if `chip_x/chip_y` should maybe omitted from the output or be called `chip_x_while_the_asol_happens_to_go_through_the_center_of_the_dither_pattern`. A similar question applies to `ccd_id`, though of course it's only a minority of sources that dithers between two chips.

kglotfelty:
Short answer: they are the "mean" chip coordinates, in so much as the `_PNT` keywords represent the "mean" pointing.  The same is true for most of the coords : theta, phi, detx, dety. They all represent the "conceptual" mean value. 

To get the time history then RA,DEC in the asol files -> RA_PNT, DEC_PNT in the keyword list.

If you have an event list then you should start with chip coords, locate the asol record at the time of the event(*), and then use those as the 'PNT' ....that's a whole different routine.  (*) foo_p_e use the last asol record with the time before the event's time.  [The handling of the data before the 1st asol record varies.]

hamogu:
In this particular application, I'm looking at one of the observations with a large dither on HRC-S, that dithers the source from the thick to the thin UV/ion shield, so I want to know, for each timestep, where the source RA/DEC fall  on the UV/ion shield (i.e. chipx/chipx)

I'll write that for my use case and then I'll see if it makes sense to generalize that. If so, I'll open a PR later; thus I'm leaving this issue open, though my immediate question is answered.

PR #976, opened 2025-08-14
Improve the coords.chandra documentation

DougBurke:
- add `get_keyword_arguments` routine to grab the keywords we need from a file

  **NOTE** are these the correct keywords as I've seen some confusion when I was playing around

- try to add documentation to try and address #954 

  **NOTE** I am 99.99% sure this documentation is incomplete; e.g. I need to add info for the individual routines and I am not convinced the examples I have added are useful to anyone but me.

### DIFF
--- a/coords/chandra.py
+++ b/coords/chandra.py
@@ -22,8 +22,8 @@
 
 The conversion between the various coordinate systems on Chandra
 depend on the aspect solution, since this determines where the
-spacecraft is pointing at any point during the observation. However,
-the routines here are only for a single point in the aspect solution,
+spacecraft is pointing at any given time during the observation. However,
+the routines here are only for one specific pointing
 given by the RA_PNT, DEC_PNT, and ROLL_PNT argument values. The
 aspect solution file can be used to find the instantaneous
 values for these fields in the "ra", "dec", and "roll" columns.


coords/chandra.py
```diff
@@ -18,17 +18,93 @@
...
-# KJG: 2016-09013 no changes needed for P3
-
-
-"""
-Routines related specifically to converting to/from Chandra coordinates
+"""Routines related specifically to converting to/from Chandra coordinates.
+
+The conversion between the various coordinate systems on Chandra
+depend on the aspect solution, since this determines where the
+spacecraft is pointing at any point during the observation. However,
```
hamogu, line 25:
```suggestion
spacecraft is pointing at any given time during the observation. However,
```
"pointing at any point" sounds somewhat funny, but might be confusing to non-native speakers to have "point" once be a spatial "pointing" and once be a "point in time" within in the same sentence.

coords/chandra.py
```diff
@@ -18,17 +18,93 @@
...
+the routines here are only for a single point in the aspect solution,
```
hamogu, line 26:
```suggestion
the routines here are only for one specific pointing
```

hamogu (approved):
I know I used this module several time in the past and I always look at the code to understand how to use it. I'm almost certain some of those uses would make up a reasonable additional example, but for the life of me I cannot recall which project that was on. 
So, I suggest to merge this, but I'll keep it in the back of my head and when I run into it again, I can always open a new PR to add an example later.

hamogu:
I'm approving because I think this is an enhancement in general, but I do have some suggestion on wording I suggest to consider before merging.

kglotfelty (approved):
I thought we already had a routine that grabbed the keywords but I can't think of where it might be ... so it's probably even more of a reason we should have a simple helper here too. 

Docs look good.

PR #989, opened 2025-08-21
cxcdm_wrapper: tweak for CLSTBITS keyword with numpy2

### LINKED ISSUE #979: cxcdm_wrapper : CLSTBITS keyword wrong
The "acis_clear_status_bits" keyword is wrong. It's written in the cxcdm_wrapper module.

kglotfelty:
This fixes #979 

FWIW: I would probably have just replaced all the binary OR logic and just replaced it with the actual string

```python
_cbits_str = '11111111011000000011111111000001'
```

and then avoided the `"{}".format(_cbits_str)` when the string was being written, ie just `_cbits_str`.

Tests running, will mark ready for review when complete.


kglotfelty:
No joy. breaks elsewhere. On to round 2.

PR #1007, opened 2025-11-06
dax: run callbacks when added to shapes

kglotfelty:
Feedback from AF,  needing to actually move/edit shape was confusing, should run callback when it is added to shape.


DougBurke:
I'm going to assume this doesn't need documentation (either in the changelog or in the dax ahelp file(s)).

PR #1031, opened 2026-02-26
Sherpa function to facilitate loading HETG responses associated with a PHA2 file

dprincipe2:
If users generated or obtained an HETG PHA2 file through relatively standard means (e.g., Chandra archive or running chandra_repro) then this function will load their PHA2 file into sherpa and find/assign the associated ARFs and RMFs.

### DIFF
--- a/sherpa_contrib/load_hetg_resp.py
+++ b/sherpa_contrib/load_hetg_resp.py
@@ -2,18 +2,12 @@ import numpy as np
 import glob
 from pathlib import Path
 from sherpa.astro.ui import load_data, load_arf, load_rmf
-from pycrates import (
-    read_file,
-    get_keyval,
-    read_pha,
-    get_history_records,
-    is_pha_type2
-)
+from pycrates import read_file, get_keyval, read_pha, get_history_records, is_pha_type2
 
 
 def find_resp_files(pha2_file_par, resp_type_par, resp_dir_par=None):
     """
-    Identifies ARFs and RMFs that are associated with an HETG PHA2 file.  
+    Identifies ARFs and RMFs that are associated with an HETG PHA2 file.
 
     Parameters
     ----------
@@ -43,12 +37,12 @@ def find_resp_files(pha2_file_par, resp_type_par, resp_dir_par=None):
 
     # load the pha2 file and obtain the crate where relevant information is stored
     pha2_dataset = read_pha(pha2_file_par)
-    
-    #Return Error if file is not a PHA2 file
+
+    # Return Error if file is not a PHA2 file
     if is_pha_type2(pha2_dataset) != True:
         raise ValueError("Input file must be of type PHA2.")
 
-    #grab the spectrum crate to determine tg_m and various header values
+    # grab the spectrum crate to determine tg_m and various header values
     spec_crate_par = pha2_dataset.get_crate(2)
 
     # identify the number of spectra in the PHA2 file
@@ -59,24 +53,24 @@ def find_resp_files(pha2_file_par, resp_type_par, resp_dir_par=None):
 
     if "Version DS" in creator_key:
 
-        # if users pha2_file_par is a long path or a single file, strip the name and check for the root so the resp 
+        # if users pha2_file_par is a long path or a single file, strip the name and check for the root so the resp
         # glob doesn't get any extra unrelated files in the response dir.
 
         # Note, the strings in this section are based on the DS (CXC Data Systems Group) standard naming.
         pha_name = Path(pha2_file_par).name
         pha_root = pha_name.partition("_pha2.fits")[0]
 
-        #uses provided resp_dir to check for files. First check for ARF/RMF.fits and if not found then check compressed.
+        # uses provided resp_dir to check for files. First check for ARF/RMF.fits and if not found then check compressed.
         if resp_dir_par != None:  # use provided resp_dir_par
             resp_list_par = glob.glob(
                 f"{resp_dir_par}/{pha_root}*{resp_type_par}*.fits"
             )
             if len(resp_list_par) == 0:
                 resp_list_par = glob.glob(
-                    f"{resp_dir_par}/{pha_root}*{resp_type_par}*.fits*" #will grab .gz files
-            )                
+                    f"{resp_dir_par}/{pha_root}*{resp_type_par}*.fits*"  # will grab .gz files
+                )
 
-        #if user does not provide a response directory
+        # if user does not provide a response directory
         else:
             pha_dir = Path(
                 pha2_file_par
@@ -86,8 +80,8 @@ def find_resp_files(pha2_file_par, resp_type_par, resp_dir_par=None):
             )
             if len(resp_list_par) == 0:
                 resp_list_par = glob.glob(
-                    f"{pha_dir}/responses/{pha_root}*{resp_type_par}*.fits*" #will grab .gz files
-                )                
+                    f"{pha_dir}/responses/{pha_root}*{resp_type_par}*.fits*"  # will grab .gz files
+                )
 
     elif (
         "Version CIAO" in creator_key
@@ -97,7 +91,7 @@ def find_resp_files(pha2_file_par, resp_type_par, resp_dir_par=None):
         hist = get_history_records(spec_crate_par)
         pha_root = ""  # set to "" (blank) so it still works for TGCAT downloaded data which has no root par.
 
-        # if chandra_repro was used to produce the PHA2 file then a ':root" value is saved in the header history. 
+        # if chandra_repro was used to produce the PHA2 file then a ':root" value is saved in the header history.
         # This identifies that value.
         for i in range(0, len(hist)):
             if ":root=" in hist[i][1]:  # find the line where the :root command was used
@@ -109,34 +103,34 @@ def find_resp_files(pha2_file_par, resp_type_par, resp_dir_par=None):
                 ]  # remove everything after the root value
                 break  # stops searching hist for the appropriate line after it is found
 
-        #if pha_root is not set then it means the file came from a tgcat downloaded PHA2 file.
+        # if pha_root is not set then it means the file came from a tgcat downloaded PHA2 file.
 
         # use glob and pha_root to find the responses
         if resp_dir_par != None:  # use provided resp_dir_par
             resp_list_par = glob.glob(f"{resp_dir_par}/{pha_root}*.{resp_type_par}")
             if len(resp_list_par) == 0:
-                resp_list_par = glob.glob(f"{resp_dir_par}/{pha_root}*.{resp_type_par}*") #will grab .gz files
+                resp_list_par = glob.glob(
+                    f"{resp_dir_par}/{pha_root}*.{resp_type_par}*"
+                )  # will grab .gz files
 
-        #if user does not provide response directory
+        # if user does not provide response directory
         else:
             pha_dir = Path(
                 pha2_file_par
             ).parent  # need to get directory where PHA2 file is located
-            resp_list_par = glob.glob(
-                f"{pha_dir}/tg/{pha_root}*.{resp_type_par}"
-            )  
+            resp_list_par = glob.glob(f"{pha_dir}/tg/{pha_root}*.{resp_type_par}")
             if len(resp_list_par) == 0:
                 resp_list_par = glob.glob(
-                    f"{pha_dir}/tg/{pha_root}*.{resp_type_par}*" #will grab .gz files
-                )                  
-            #TGCAT does not come with a tg directory so check in the pha_dir for response files.
+                    f"{pha_dir}/tg/{pha_root}*.{resp_type_par}*"  # will grab .gz files
+                )
+            # TGCAT does not come with a tg directory so check in the pha_dir for response files.
             if len(resp_list_par) == 0:
                 resp_list_par = glob.glob(
-                    f"{pha_dir}/{pha_root}*.{resp_type_par}" #will grab .gz files
-                )            
+                    f"{pha_dir}/{pha_root}*.{resp_type_par}"  
+                )
             if len(resp_list_par) == 0:
                 resp_list_par = glob.glob(
-                    f"{pha_dir}/{pha_root}*.{resp_type_par}*" #will grab .gz files
+                    f"{pha_dir}/{pha_root}*.{resp_type_par}*"  # will grab .gz files
                 )
 
     # if the creator of the PHA2 file is not DS or CIAO then raise error and exit.
@@ -151,9 +145,11 @@ def find_resp_files(pha2_file_par, resp_type_par, resp_dir_par=None):
             "Could not identify responses. Please load responses manually."
         )
 
-    #Some response files were found so tell user matching will begin
-    print(f'\n{resp_type_par.upper()} files identified. Attempting to match them to PHA2 spectra:')
-    print('-'*63)
+    # Some response files were found so tell user matching will begin
+    print(
+        f"\n{resp_type_par.upper()} files identified. Attempting to match them to PHA2 spectra:"
+    )
+    print("-" * 63)
 
     # check that the length of the ARF or RMF lists match the number of PHA spec in the PHA2 file
     if len(resp_list_par) != num_spec_par:
@@ -213,34 +209,36 @@ def match_resp_order(pha2_file_par, resp_list_par, resp_type_par):
     resp_pha2_tg_part_arr = []
     resp_obsid_arr = []
 
-    #start a counter to determine how many response files are missing an OBS_ID header value
+    # start a counter to determine how many response files are missing an OBS_ID header value
     obsid_missing_count = 0
 
     # read each response file and append appropriate header values
-    for i in range(0, len(resp_list_par)):
-        resp_data = read_file(resp_list_par[i])
+    for i in resp_list_par:
+        resp_data = read_file(i)
         resp_m_arr.append(get_keyval(resp_data, "TG_M"))
         resp_pha2_tg_part_arr.append(get_keyval(resp_data, "TG_PART"))
-        
-        #As of 3/23/26, TGCAT-provided RMFs do not have a 'OBS_ID' header keyword. This attempts to mitigate the issue.
-        #If there is any error reading the OBS_ID header then it assigns the PHA2 obsid value to the array and prints a
+
+        # As of 3/23/26, TGCAT-provided RMFs do not have a 'OBS_ID' header keyword. This attempts to mitigate the issue.
+        # If there is any error reading the OBS_ID header then it assigns the PHA2 obsid value to the array and prints a
         # warning  so the rest of code will continue. TGCAT RMF data is expected to get OBSID par in near future.
         try:
             resp_obsid_arr.append(get_keyval(resp_data, "OBS_ID"))
         except:
             resp_obsid_arr.append(pha2_tg_obsid)
-            obsid_missing_count = obsid_missing_count+1
+            obsid_missing_count = obsid_missing_count + 1
 
-    #if any number of response files don't have an obsID parameter then warn the user this wont be used to check. Still
-    #continue though because this requirement is likely to almost never be an issue.
+    # if any number of response files don't have an obsID parameter then warn the user this wont be used to check. Still
+    # continue though because this requirement is likely to almost never be an issue.
     if obsid_missing_count > 0:
-        print(f'\nWARNING -- Header keyword OBS_ID is missing from {obsid_missing_count} {resp_type_par.upper()} file(s). {resp_type_par.upper()}(s) will not be checked against obsID for validity. \n')
+        print(
+            f"\nWARNING -- Header keyword OBS_ID is missing from {obsid_missing_count} {resp_type_par.upper()} file(s). {resp_type_par.upper()}(s) will not be checked against obsID for validity. \n"
+        )
 
     # convert obsid lists to numpy arrays for later use with np.where() for obsID checking
     pha2_tg_obsid = np.array(pha2_tg_obsid)
     resp_obsid_arr = np.array(resp_obsid_arr)
 
-    # create an empty object array the same size as the PHA2 file (number of spectra) to hold either 'no match' or the 
+    # create an empty object array the same size as the PHA2 file (number of spectra) to hold either 'no match' or the
     # path to the matched response file
     matched_resp_list_par = np.array([""] * num_spec_pha2, dtype="object")
 
@@ -267,7 +265,9 @@ def match_resp_order(pha2_file_par, resp_list_par, resp_type_par):
             )
 
     # report the files matched to the screen in a nice format so it is clear it worked or didn't work
-    print(f"\nThe following {resp_type_par.upper()} response files were found for obsID {pha2_tg_obsid}\n")
+    print(
+        f"\nThe following {resp_type_par.upper()} response files were found for obsID {pha2_tg_obsid}\n"
+    )
 
     # name the arm and order something more readable for output message
     arm = lambda x: "HEG" if x == 1 else "MEG" if x == 2 else "ERROR"
@@ -280,9 +280,11 @@ def match_resp_order(pha2_file_par, resp_list_par, resp_type_par):
         )
     print()
 
-    #check if no responses matched and report to user. Most common issue would be the OBSID parameter is wrong
-    if (matched_resp_list_par == 'no match').all():
-        print(f'\nWARNING -- No {resp_type_par.upper()} files matched the PHA2 spectra. Check that the ARFS and RMFs are from the same obsID\n')
+    # check if no responses matched and report to user. Most common issue would be the OBSID parameter is wrong
+    if (matched_resp_list_par == "no match").all():
+        print(
+            f"\nWARNING -- No {resp_type_par.upper()} files matched the PHA2 spectra. Check that the ARFS and RMFs are from the same obsID\n"
+        )
 
     # returns the array of matched responses in the same order as the PHA2 spectra
     return matched_resp_list_par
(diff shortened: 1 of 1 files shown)


sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,370 @@
+import numpy as np
+import glob
+from pathlib import Path
+from sherpa.astro.ui import load_data, load_arf, load_rmf
+from pycrates import read_file, get_keyval, read_pha, get_history_records, is_pha_type2
+
+
+def find_resp_files(pha2_file_par, resp_type_par, resp_dir_par=None):
+    """
+    Identifies ARFs and RMFs that are associated with an HETG PHA2 file.
```
hamogu: Doc string should start with one line of text (summary), then a blank line, then more text (if required)

```suggestion
    """Identifies ARFs and RMFs that are associated with an HETG PHA2 file.
```

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,370 @@
...
+        PHA2 spectrum which is typically provided in the chandra archive of a downloaded HETG observation or after
+        running chandra_repro on a chandra HETG obsID.
+    resp_type_par: 'arf' or 'rmf'
+        The type of response files for matchign to PHA spectra
+    resp_dir_par: directory
+        The directory where the HETG response files associated with the PHA2 file are located. If none provided, it
+        will attempt to search for them.
+
+    Returns
+    ----------
```
hamogu:
```suggestion
    Returns
    --------
```

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,370 @@
...
+    resp_list_par: list
+    Returns a list of ARF or RMF files found.
```
hamogu:
```suggestion
        Returns a list of ARF or RMF files found.
```

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,370 @@
...
+    Identifies ARFs and RMFs that are associated with an HETG PHA2 file.
+
+    Parameters
+    ----------
+
+    pha2_file_par : PHA2 fits file
+        PHA2 spectrum which is typically provided in the chandra archive of a downloaded HETG observation or after
+        running chandra_repro on a chandra HETG obsID.
+    resp_type_par: 'arf' or 'rmf'
+        The type of response files for matchign to PHA spectra
```
hamogu:
```suggestion
        The type of response files for matching to PHA spectra
```

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,370 @@
...
+
+        # if users pha2_file_par is a long path or a single file, strip the name and check for the root so the resp
+        # glob doesn't get any extra unrelated files in the response dir.
+
+        # Note, the strings in this section are based on the DS (CXC Data Systems Group) standard naming.
+        pha_name = Path(pha2_file_par).name
+        pha_root = pha_name.partition("_pha2.fits")[0]
+
+        # uses provided resp_dir to check for files. First check for ARF/RMF.fits and if not found then check compressed.
+        if resp_dir_par != None:  # use provided resp_dir_par
```
hamogu: The recommended way in Python to check if someone is None is to use "is" or "is not"
```suggestion
        if resp_dir_par is not None:  # use provided resp_dir_par
```

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,370 @@
...
+            resp_list_par = glob.glob(
+                f"{resp_dir_par}/{pha_root}*{resp_type_par}*.fits"
+            )
+            if len(resp_list_par) == 0:
+                resp_list_par = glob.glob(
+                    f"{resp_dir_par}/{pha_root}*{resp_type_par}*.fits*"  # will grab .gz files
```
hamogu: If we re only looking for gz files, while not be explicit?
```suggestion
                    f"{resp_dir_par}/{pha_root}*{resp_type_par}*.fits.gz"
```
I understand that we aim to make this work for standard file structures and that it won't work for cases where users have renamed or moved files, but I've used names like "XXX.arf.fits_old" when I tried different settings for creating an ARF. And in this case, I can't see how a more stringent check would hurt.

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,370 @@
...
+        else:
+            pha_dir = Path(
+                pha2_file_par
+            ).parent  # need to get directory where PHA2 file is located
+            resp_list_par = glob.glob(
+                f"{pha_dir}/responses/{pha_root}*{resp_type_par}*.fits"
+            )
+            if len(resp_list_par) == 0:
+                resp_list_par = glob.glob(
+                    f"{pha_dir}/responses/{pha_root}*{resp_type_par}*.fits*"  # will grab .gz files
```
hamogu:
```suggestion
                    f"{pha_dir}/responses/{pha_root}*{resp_type_par}*.fits.gz"
```

hamogu: Same is true below, but I stop commenting here. Maybe there is a good reasons @dprincipe2 didn't do that that I'm missing, but otherwise I suggest to use the same, explicit pattern everywhere.

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,370 @@
...
+    if len(resp_list_par) == 0:
+        raise ValueError(
+            "Could not identify responses. Please load responses manually."
+        )
+
+    # Some response files were found so tell user matching will begin
+    print(
+        f"\n{resp_type_par.upper()} files identified. Attempting to match them to PHA2 spectra:"
+    )
+    print("-" * 63)
```
hamogu: I don't think we need output like this. This isn't going to take to long that a user is wondering what's going on and it can be annoying to just rad a file and get too many lines of print output that I can't turn off.

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,370 @@
...
+
+    """
+
+    # check to make sure responses are either arf or rmf
+    if resp_type_par != "arf" and resp_type_par != "rmf":
+        raise ValueError("Response type must be either arf or rmf")
+
+    # reads in the pha2 file and checks how many spectra (orders) it includes.
+    pha2_dataset = read_pha(pha2_file_par)
+    spec_crate_par = pha2_dataset.get_crate(2)
```
hamogu: Generally, it's safer to select FITS extensions by name, because the order is usually not fixed in specs. However, in practice, we are only loading Chandra generated ARFs and RMFs, and for those it will always be in that same location, so in this case, we don't have to change this line.

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,370 @@
...
+        elif len(match_resp) > 1:
+            raise ValueError(
+                f"ERROR, more than one {resp_type_par.upper()} match found for TG_M={pha2_tg_m_arr[i]}, TG_PART={pha2_tg_part_arr[i]} and obsID={pha2_tg_obsid}. "
+            )
+        elif len(match_resp) == 0:
+            matched_resp_list_par[i] = "no match"
+        else:
+            raise ValueError(
+                f"ERROR - Something with wrong identifying {resp_type_par.upper()}s for TG_M={pha2_tg_m_arr[i]}, TG_PART={pha2_tg_part_arr[i]} and obsID={pha2_tg_obsid}"
+            )
```
hamogu: This `else` block can never be triggered. Mathematically, `len(match_resp) is either 0 or 1 or >1, so it will never reach the else clause.
```suggestion
```

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,370 @@
...
+
+    # report the files matched to the screen in a nice format so it is clear it worked or didn't work
+    print(
+        f"\nThe following {resp_type_par.upper()} response files were found for obsID {pha2_tg_obsid}\n"
+    )
+
+    # name the arm and order something more readable for output message
+    arm = lambda x: "HEG" if x == 1 else "MEG" if x == 2 else "ERROR"
+    order = lambda x: f"+{x}" if x > 0 else f"-{-1*x}" if x < 0 else "ERROR"
```
hamogu: It is [generally considered not good practice](https://www.flake8rules.com/rules/E731.html) to assign lambda functions to a variable. Personally, I don't care that much, but in this particular case, it's better to be even simpler: `arm = {1: 'HEG', 2: 'MEG'}` is more readable, shorter, can be be used below as `f"{arm[pha2_tg_part_arr[i]] ... `.

The `order` function doesn't need a a definition at all, but just standard Python number formatting: `f'{pha2_tg_m_arr[i]:+}'` 
So, together the print line would read:
```
        print(
            f"{arm[pha2_tg_part_arr[i]]}{{pha2_tg_m_arr[i]:+}} -- {resp_type_par.upper()}: {matched_resp_list_par[i]}"
        )
```

sherpa_contrib/load_hetg_resp.py
hamogu: Same is true below, but I'm just commenting here once.

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,370 @@
...
+    search for the responses in subdirectories where the PHA2 file is located using standard CIAO directory naming
+    formats.
+
+    Example: load_hetg_pha2(pha2_file=16370/repro/acisf16370_repro_pha2.fits, dataset_id_start=1)
+             load_hetg_pha2(pha2_file=16371/repro/acisf16371_repro_pha2.fits, dataset_id_start=13)
+
+    Parameters
+    ----------
+
+    pha2_file : PHA2 fits file
```
hamogu:
```suggestion
    pha2_file : str
```

hamogu (changes requested):
In general, I think a contribution for this functionality is very useful. Yes, users *can* (and do) load PHA2 files without it, but it's a pain to do so. It's just the type of quality-of-life issue that might make people use or not use a specific spectral fitting program.

When reading this, what struck me was how complicated the logic is to try out all the common designs because there are just so many options and how easy it would be to make the algorithm fail by just renaming a file or having additional files in that directory that accidentally match the name[*] (say, from extracting multiple sources for the same OBID with different ARFs or trying out different settings for the ARFs that generated). 
You can argue that that's beyond the scope of this PR, but I wonder if the following methods would be easier and at the same time more robust: Make use of the meta information in the header. We don't need the file name to decide if it's an ARF or an RMF. That's all specified in the FITS header (ARFs use "HDUCLAS1=RESPONSE" "HDUCLAS2=SPECRESP", RMFs use "HDUCLAS1=RESPONSE" and "HDUCLAS2=RSP_MATRIX" with "HDUCLAS3=REDIST" in the scheme that Chandra uses). So, in this, we would ask for a root directory, try to load every file as a fits file independent of the same. We catch the errors, and simply move to the next file. For every file that opens we check the fits header to get the file type, the ObsId, and all the other information that the code needs. If we don't find anything, we can go to subdirectories and try those (at least those with commonly used names like in the current PR).
In that way, we don't miss files that are named in unusual ways and we are not deterred by having other random files in the same directory unless those files are also valid ARFs and RMFs in the same directory for the same instrument and obsid.
That would require some rewriting but won't be hard, and I believe it might be worth the effort.
On the other hand, this approach fails more easily and thus avoids matching the wrong arf and rmf (that maybe shouldn't be in that directory) accidentally.
@DougBurke @dprincipe2 Thoughts?

I've added in-line comments about small things but of course those are only relevant if we go ahead with the approach currently taken in the PR.

*: That can be avoided by having clean directories and good file naming and good naming root, e.g. if I have "star_flare.arf" and "star_out_of_flare.arf" and can't just pass in "star" as root.

dprincipe2:
Thanks for your comments @hamogu. Most of the individual line changes you suggest I think are fair and I can update them. As for the higher level comments, these functions are to cover the majority of cases for users loading HETG responses. As you suggest, I do think it would be more difficult to get something that will work 100% of the time. That being said, I think this will work for probably ~90% of situations. Very high level HETG users can probably live without something like this if necessary. 

This is designed to work for HETG data (a) downloaded from chaser/archive, (b) downloaded from tgcat and (c) created when running chandra_repro with the default settings OR chandra_repro extractions on other sources in the field. I haven't checked (but maybe I should) to see if it would work for tgextract data. 

One thing I think you overlooked is that this uses the 'root' name from the PHA2 file header history to identify corresponding ARFs/RMFs. So your example of extracting multiple sources from the same obsID will actually work as long as the user doesn't manually change the name of the PHA2 file after it was created. If there are four different extractions in the same directory where different names are chosen when running chandra_repro with the 'root' parameter then this code will correctly identify the associated responses without (incorrectly) identifying the wrong ones. 

One minor issue I had was the RMFs produce by tgcat currently don't have an OBSID header keyword. This will change in the future when tgcat is eventually migrated to something newer than CIAO 415 (or whatever old version it is using) but for now, I can't use the OBSID to double check that it grabbed the correct RMF. I think thats only a minor issue. 

@hamogu @DougBurke , if you haven't yet, I would encourage you to try the function and let me know if you run into some of the issues you think would be detrimental to a significant group of Chandra users. All I know is that loading responses is currently really annoying for HETG and I suspect a lot of users, probably specifically new users of HETG, will hit this roadblock and get frustrated. I believe this implementation will be much better than nothing. If you wanted to re-do this from scratch or update this in a later release that would work too. I think we should get something like this out to users in some form though.

sherpa_contrib/load_hetg_resp.py
dprincipe2: Yeah, we can remove this or put it behind a verbose setting or something. I personally like to double check things were found correctly but it is definitely a lot of text to print to the screen. If this works for most cases, which I think it does, the text is more of a burden.

DougBurke:
There is an argument that perfect is the enemy of good, so if this works for the common cases that we can take this and then re-work at a later date to be more-generic if needed. As a non-gratings person I think I'd like to know the common ways people will have created/accessed PHA2 + responses and see if it works for "standard" situations (e.g. if a user has manually created files then maybe we don't expect this to work?)

- `TGCat`
- `chandra_repro` (we just got the `tgorders` option added)
- `tgextract` + `fullgarf`
  - what other ways are there (e.g. do we care abut `combine_grating_spectra`
  - presumably `crosscrissssss`

DougBurke:
Just to note that DPH asked about LETG support. Is it reasonable here, or in a later release or ?

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,370 @@
...
+
+    This function loads an HETG PHA2 file and any associated ARF and RMF response files. If matching responses are
+    found only for a subset of HETG orders (e.g., orders +1 and -1) then only those order's responses will be loaded.
+    This works only for PHA2 files and not PHA files. If arf_dir or rmf_dir are not provided then this tool will
+    search for the responses in subdirectories where the PHA2 file is located using standard CIAO directory naming
+    formats.
+
+    Example: load_hetg_pha2(pha2_file=16370/repro/acisf16370_repro_pha2.fits, dataset_id_start=1)
+             load_hetg_pha2(pha2_file=16371/repro/acisf16371_repro_pha2.fits, dataset_id_start=13)
+
```
hamogu: file names must be quated since they are strings
```suggestion
    Example
    -------
    >>> load_hetg_pha2(pha2_file="16370/repro/acisf16370_repro_pha2.fits", dataset_id_start=1)
    >>> load_hetg_pha2("pha2_file=16371/repro/acisf16371_repro_pha2.fits", dataset_id_start=13)

```

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,370 @@
...
+    # for every identified response (arf), load matching arf and rmf. It should be ok to load RMFs with the arf loop
+    # cause an error would be previously thrown if every arf didn't have a matching rmf.
+    for i in range(len(arf_list_matched)):
+        if arf_list_matched[i] != "no match":
+            # note, i+dataset_id_start here cause sherpa starts with dataset 1 and not 0.
+            load_arf(i + dataset_id_start, arf_list_matched[i])
+            load_rmf(i + dataset_id_start, rmf_list_matched[i])
+
+    return ()
```
hamogu: Don't return an empty tuple. It just prints and extra line "()" on my screen. Either delete the return line (it's not needed) or, if you really want it, `return None` (which is the same as not having the line)

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,370 @@
...
+        # if chandra_repro was used to produce the PHA2 file then a ':root" value is saved in the header history.
+        # This identifies that value.
+        for i in range(0, len(hist)):
+            if ":root=" in hist[i][1]:  # find the line where the :root command was used
+                root_line = (
+                    hist[i][1].split("=", 1)[1].strip()
+                )  # strip the unnecessary stuff but leaves the extra spaces
+                pha_root = root_line.split(" ")[
+                    0
+                ]  # remove everything after the root value
```
hamogu, line 103: If I use a rootname that's longer than ~60 characters, it can't find the files:
```
chandra_repro root=a_very_long_String_that_probabl_doesnt_happen_in_real_live_but_Ill_try
```
and then
```
load_hetg_pha2(
    pha2_file="/Users/guenther/data/Chandra/5/repro2/a_very_long_String_that_probabl_doesnt_happen_in_real_live_but_Ill_try_repro_pha2.fits",
    dataset_id_start=1,
)
```
gives mw an error message:
```
ValueError: Could not identify responses. Please load responses manually.
```
That's because root name doesn't fit into a single card and it broken over two cards. In principle, that's fixable by merging the continuing card, but I don't think it's going to happen in practice, so we shouldn't worry about it.

hamogu (changes requested):
I tried is on a couple of spectra that I happened to have on disk and it worked every time. 
I'll leave it to @dprincipe2 to decide if he add LEGT (and possibly HRC) here or in a future version.

dprincipe2:
> Just to note that DPH asked about LETG support. Is it reasonable here, or in a later release or ?

Looks like the LETG+ACIS-S datasets already work with the function as is. LETG+HRC-S has some issues and I will take a closer look to see if it's easy to do before the first release. We'll have to rename the function to something like load_grating_pha2() since it at least works with LETG+ACIS.

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,372 @@
+import numpy as np
+import glob
+from pathlib import Path
+from sherpa.astro.ui import load_data, load_arf, load_rmf
+from pycrates import read_file, get_keyval, read_pha, get_history_records, is_pha_type2
+
+
+def find_resp_files(pha2_file_par, resp_type_par, resp_dir_par=None):
```
hamogu: Can we not call every parameter `*_par`? If they are in the function signature, they are parameters.

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,372 @@
...
+    # check that the length of the ARF or RMF lists match the number of PHA spec in the PHA2 file
+    if len(resp_list_par) != num_spec_par:
+        print(
+            f"\nWARNING-- The identified number of {resp_type_par.upper()}s [{len(resp_list_par)}] does not match the number of PHA spectra [{num_spec_par}] in the PHA2 file. Only the responses found will be included.\n"
+        )
+
+    return resp_list_par
+
+
+def match_resp_order(pha2_file_par, resp_list_par, resp_type_par, verbose_par=False):
```
hamogu: The parameter is called `verbose` in all of CIAO and in all contributed scripts. Can we use that name here, too, instead of making this is one and only case where it's called different (`verbose_par`)?

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,372 @@
...
+    if (matched_resp_list_par == "no match").all():
+        print(
+            f"\nWARNING -- No {resp_type_par.upper()} files matched the PHA2 spectra. Check that the ARFS and RMFs are from the same obsID\n"
+        )
+
+    # returns the array of matched responses in the same order as the PHA2 spectra
+    return matched_resp_list_par
+
+
+def load_hetg_pha2(pha2_file=None, arf_dir=None, rmf_dir=None, dataset_id_start=1, use_errors_par=False, verbose=False):
```
hamogu: Same here for `use_errors_par`. It's literally just passed through to the PHA files reading, where it's called `user_errors`. Please use the same name, and drop the `_par` part.

hamogu (changes requested):

dprincipe2:
Ok, I think this is good to go! The only thing someone may want to take a closer look at is the new try statements I added to handle TGCat-provided RMFs missing some keywords that the archive/chandra_repro RMFs have. This now works with all gratings (HETG + LETG-ACIS/HRC) so I renamed it to load_gratings_pha2(). 

@hamogu @DougBurke

sherpa_contrib/load_hetg_resp.py
```diff
@@ -5,90 +5,90 @@
...
+    Identifies ARFs and RMFs that are associated with an HETG/LETG PHA2 file.
 
     Parameters
     ----------
 
-    pha2_file_par : PHA2 fits file
-        PHA2 spectrum which is typically provided in the chandra archive of a downloaded HETG observation or after
-        running chandra_repro on a chandra HETG obsID.
-    resp_type_par: 'arf' or 'rmf'
+    pha2_file : PHA2 fits file
```
hamogu: The input type is "str" here, right? Or is it an open file handle? I think that's what we want to write here. The fact that its a PHA2 file can go into the description in the next line.

sherpa_contrib/load_hetg_resp.py
```diff
@@ -5,90 +5,90 @@
...
+        PHA2 spectrum which is typically provided in the chandra archive of a downloaded /LETG observation or after
+        running chandra_repro on a chandra HETG/LETG obsID.
+    resp_type: 'arf' or 'rmf'
         The type of response files for matching to PHA spectra
-    resp_dir_par: directory
-        The directory where the HETG response files associated with the PHA2 file are located. If none provided, it
+    resp_dir: directory
```
hamogu: String? `pathlib.Path` object? I assume "str", so we should say that. The fact that the string is the name off a directory can go into the line below in the description.

sherpa_contrib/load_hetg_resp.py
```diff
@@ -140,65 +141,65 @@ def find_resp_files(pha2_file_par, resp_type_par, resp_dir_par=None):
...
     find_resp_files() and put them in order of the PHA2 spectra.
 
     Parameters
     ----------
 
-    pha2_file_par : PHA2 fits file
-        PHA2 spectrum which is typically provided in the chandra archive of a downloaded HETG observation or after
-        running chandra_repro on a chandra HETG obsID.
-    resp_type_par: 'arf' or 'rmf'
+    pha2_file : PHA2 fits file
```
hamogu: Same as above. The type is "str"

sherpa_contrib/load_hetg_resp.py
```diff
@@ -140,65 +141,65 @@ def find_resp_files(pha2_file_par, resp_type_par, resp_dir_par=None):
...
+        PHA2 spectrum which is typically provided in the chandra archive of a downloaded HETG/LETG observation or after
+        running chandra_repro on a chandra HETG/LETG obsID.
+    resp_type: 'arf' or 'rmf'
         The type of response files for matchign to PHA spectra
-    resp_dir_par: directory
-        The directory where the HETG response files associated with the PHA2 file are located. If none provided, it
+    resp_dir: directory
```
hamogu: str

sherpa_contrib/load_hetg_resp.py
```diff
@@ -211,15 +212,34 @@ def match_resp_order(pha2_file_par, resp_list_par, resp_type_par, verbose_par=Fa
...
+        #identify the gratings type (HEG, MEG or LEG).
+        try:
+            resp_pha2_tg_part_arr.append(get_keyval(resp_data, "TG_PART"))
+        except:
+            grating_type = get_keyval(resp_data, "GRATING")
+            if grating_type == 'LETG':
+               resp_pha2_tg_part_arr.append(3) #note LEG --> tg_part = 3
+            else:
+                raise ValueError(
+                    f"ERROR-- Could not identify grating type and/or order. Please check load responses manually.")
```
hamogu: word missing: check AND load?

hamogu (changes requested):
I think this is getting close. As you cna see, my comments are getting smaller!

sherpa_contrib/load_hetg_resp.py
```diff
@@ -0,0 +1,392 @@
...
+
+    matched_resp_list: array
+        Returns an array of ARF or RMF file paths matched to the order (arrangement) of the input PHA2 file. The string
+        value 'no match' is returned for elements where there is a spectrum in the PHA2 file but no matching response
+        file.
+
+    """
+
+    # check to make sure responses are either arf or rmf
+    if resp_type != "arf" and resp_type != "rmf":
```
hamogu, line 192: This is fine since there are only two options. If you had more, you could also use `if resp_type not in ['arf', 'rmf]:`

hamogu (approved):
I think the content is good. I'll leave it to @DougBurke to comment if the style is consistent enough (e.g. the docstrings start with an empty line, not a one-line summary as usual in Python).

hamogu:
I'm removing the "draft" label because I believe that this is ready - we just need @DougBurke or @kglotfelty to check that the formatting conventions (line length in comments, extend of docstrings etc. are sufficient). It would be great to merge this soon, so that it can be used in crisscross already (currently, the crisscross PR duplicates this code).

hamogu (approved):
I just rebased this on current main. Can we merge?

PR #1050, opened 2026-05-13
check_ciao_version: add packages in the LINK list to the packages needing updating in the current env

### LINKED ISSUE #1045: check_ciao_version not picking up ciao-contrib update
```
% conda activate ciao-4.18
% conda list | grep contrib
ciao-contrib                     4.18.0           py_5                 https://cxc.cfa.harvard.edu/conda/ciao
% check_ciao_caldb
% check_ciao_version
CIAO (installed via conda) is up to date.
```

I have CALDB env variables overridden so `check_ciao_version` doesn't recognize it needs updating -- that's ok, `check_ciao_caldb` does

```bash
% check_ciao_caldb -l
CALDB environment variable = /Applications/CALDB
             CALDB version = 4.12.3
             release date  = 2025-12-16T16:00:00 UTC
CALDB query completed successfully.
# check_ciao_caldb (01 July 2020): ERROR The latest available CALDB version is 4.12.4
```

kglotfelty:
This fixes #1045 

`check_ciao_version` is checking the `FETCH` list for packages that need to be updated in the current environment.  But if the package has already been installed in a separate environment and it's still in the cache, then it does not show up in the `FETCH` list; it's in the `LINK` list.  This update combines the packages listed in the `FETCH` and `LINK` lists. 

Not likely to affect most users since they only have one environment so not likely to be in cache already.

### DIFF
--- a/ciao_contrib/_tools/versioninfo.py
+++ b/ciao_contrib/_tools/versioninfo.py
@@ -444,7 +444,9 @@ def check_conda_versions(ciao: str) -> bool:
     # filter the names list to only show the CIAO packages
     #
     names = []
-    for name, version in get_names(fetch):
+    all_names = get_names(fetch)
+    all_names.extend(get_names(link))
+    for name, version in all_names:
         if name not in packages:
             v3(f" - skipping {name} {version} from FETCH list")
             continue


DougBurke:
I don't think I want to try and understand this ...

PR #1053, opened 2026-06-16
dax: include sigma in stats

kglotfelty:
From helpdesk request :  add `dmstat`'s `sigma` (standard deviation) to the statistics output.

### DIFF
--- a/share/doc/xml/dax.xml
+++ b/share/doc/xml/dax.xml
@@ -136,6 +136,16 @@ unix% ds9 -analysis $ASCDS_CONTRIB/config/ciao.ds9 ...
 
 </DESC>
 
+<ADESC title="Changes in script 4.18.2 (July 2026) release">
+  <PARA>
+        The "Statistics (dmstat)" task now reports the standard deviation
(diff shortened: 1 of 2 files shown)


share/doc/xml/dax.xml
```diff
@@ -136,6 +136,16 @@ unix% ds9 -analysis $ASCDS_CONTRIB/config/ciao.ds9 ...
 
 </DESC>
 
+<ADESC title="Changes in script 4.18.2 (July 2026) release">
+  <PARA>
+        The "Statistics (dmstat)" task now reports the standard deviation
+        of the pixel values in the regions. Note: This reports the biased standard
+        deviation estimator (divided by N)  that the dmstat tool computes.
```
DougBurke:
```suggestion
        of the pixel values in the regions. Note: This reports the value calculated
        by dmstat, which is the biased standard deviation estimator (divided by N).
```

DougBurke (commented):
stupid github "review pending" behaviour .,....

PR #1055, opened 2026-06-23
Update for the 4.18.2 release

DougBurke:
Assume that it will be in July

PR #1056, opened 2026-06-24
Document crisscross and specextract changes

DougBurke:


DougBurke:
@nplee - is the specextract change described adequately here?

DougBurke:
@hamogu - I mainlylike this file as it's useful for us to see what has gone into a release, so I do not know how many people actually read it other than me. Saying that, is the crisscross text okay (and probably the section on sherpa_contrib.load_hetg_resp)?

Also, thanks to merging #1031 - then making a minor tweak to the docstring to fix an example in #1055 - and then #1021 I just wanted to check that `sherpa_contrib.load_hetg_resp` is correct / as expected.

nplee:
@DougBurke yes, this suffices for the user; the changes shouldn't affect regular/typical workflows.

hamogu (approved):

PR #1057, opened 2026-06-24
`flux_obs` memory and open file limit checks

nplee:
This update to flux_obs is primarily linting of Python 2isms into Python 3 in the lower-level files but is driven by recent helpdesk tickets where users were unable to run `merge_obs` to completion due to running out of usable system memory and swap space, either from really large FOVs or having a large number of observations.  The failure stems from `flux_obs`, and even if this fails via `merge_obs`, the reprojection step performed is still useful and those files are retained, and the user may subsequently run `flux_obs` independently.

This update will attempt to estimate the memory usage for generating the data products and stacking the subsequent images and throw an error, either suggesting a maximum number of cores to use if run in parallel or to run the process serially; furthermore, adjusting the file chunk size passed along to he stacking routines is performed to reduce memory usage if needed.

This branch can be rebased if necessary.

### DIFF
--- a/bin/flux_obs
+++ b/bin/flux_obs
@@ -452,8 +452,8 @@ def project_mem_use(obsinfos, xygrid, ncore) -> tuple[int,int]:
 
     with fi.Project_Memory_Use(evtfiles=_evts, filesize=img_filesize, ncore=ncore) as memcheck:
         _ = memcheck.project_parallel_memory()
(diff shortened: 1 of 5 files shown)


DougBurke:
I think I'd prefer to see this rebased. Unfortunately when I tried it was not obvious what best to do, so it depends on how well you jell with your rebase skills...

nplee:
@DougBurke I think I managed to rebase the branch ::knock on wood::

PR #1061, opened 2026-06-29
chandra_repro: fix meta-data 

kglotfelty:
@nplee noticed that the `chandra_repro` `HISTORY` records was missing the `patch_hrc_ssc` parameter. It was also missing the `tg_orders` parameter.   He also noticed that the `FLTFILE` keyword was set to a temporary file name.

This fixes both issues.

Relatively minor/safe changes but I'm fine for leaving for the next release.

### DIFF
--- a/bin/chandra_repro
+++ b/bin/chandra_repro
@@ -34,7 +34,7 @@ Aim:
 """
 
 toolname = "chandra_repro"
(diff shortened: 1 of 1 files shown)


DougBurke:
I don't think these "fixes" need any sort of documentation.

PR #1062, opened 2026-06-29
`convert_xspec_script` failing on ciao-install distribution looking for $HEADAS

nplee:
`convert_xspec_script` is looking for the `$HEADAS` variable with the `binexe/python` being used in the `ciao-install` distributed CIAO environment when trying to import `sherpa.astro.xspec`.  This isn't a problem for the Conda distribution since the variable is set.  

In this modification of the script, if the `$HEADAS` variable is not present, the script sets it for the Python instance to match the `$CIAO_HEADAS` variable which is set for all CIAO distributions, presently.

### DIFF
--- a/Changes.CIAO_scripts
+++ b/Changes.CIAO_scripts
@@ -28,6 +28,11 @@ Updated Scripts
     Improve the behaviour when CIAO is installed into multiple conda
     environments.
 
+  convert_xspec_script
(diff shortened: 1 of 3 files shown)


DougBurke:
I added minimal documentation (just to remind people this script exists ;-)

PR #1065, opened 2026-06-30
Remove crisscross tutorial now hosted independently of this repository

hamogu:


setup.py
nplee, line 1: looks good; packaging routine is behaving as expected with these changes.

dprincipe2 (changes requested):
@hamogu, can you also remove the full_coup_src_list.tsv? Both these files were just used when running/testing the notebooks.

hamogu:
@dprincipe2 PR updated to also remove that file.

nplee (approved):
packaging with these revisions look good; I will go ahead and merge if no further comments by COB.

PR #1066, opened 2026-07-08
chandra_repro: update for OBC mode that omit ASPTYPE keyword in header

kglotfelty:
OBS_ID 1304 does not even have `ASPTYPE` keyword so add defensive check to force down the OBC path.

### DIFF
--- a/bin/chandra_repro
+++ b/bin/chandra_repro
@@ -34,7 +34,7 @@ Aim:
 """
 
 toolname = "chandra_repro"
-version = "29 June 2026"
+version = "08 July 2026"
 
 # import standard python modules as required
 #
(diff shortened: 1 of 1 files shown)


DougBurke:
@kglotfelty Do you have an idea of how many other obsids are there like this (I assume only a handful)?

DougBurke:
[I'm assuming we should take this for the contrib update, as we are still waiting for some doc fixes and this is a highly-localised fix?]

kglotfelty:
There look like there are 4 total in the "science" OBS_ID range ; that is there are bunch with OBS_ID > 40k but those are engineering obsids.  These 4 are also really engineering obsids, but that's early in the mission and we were not distinguishing them clearly.  

There's no rush IMO.  If there's going to be another tar file built then sure, but definitely not a driver.

nplee:
there will be another tarball, so might as well; it'll motivate me to modify the build script sooner rather than later while the tarball modifications are still lingering around in some brain crevice.

DougBurke:
I don't think it's worth a documentation note.

PR #1068, opened 2026-07-08
Updated several crisscross-related files with small fixes for compati…

dprincipe2:
…bility with CIAO.

--modified crisscross executable to convert ciao parameter string 'None' values to bool None. This bug prevented  'single_src_pos' and 'single_src_root' parameters from being set properly.

--modified crisscross.xml and cleanspec.xml to fix 'Examples' text  so they now work if copy and pasted into a terminal.

--modified 'file exists' error statement in crisscross.py so it's more consistent with clobber=yes ciao nomenclature.

--removed comments at top of criss_cross.py about future improvements.

### DIFF
--- a/ciao_contrib/criss_cross/criss_cross.py
+++ b/ciao_contrib/criss_cross/criss_cross.py
@@ -15,12 +15,7 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
-# List of things I still need to do:
-# - add back in ds9 figure generation so users can see where confusion occurs on the evt2 fits image.
-# - add functionality for user to calculate confusion between two selected sources in a field of view and print results to screen.
-# - Write parameters values used to run into header (either as history or as keywords)
-##########################################################################################
-##########################################################################################
+
 ##########################################################################################
 import glob
 import os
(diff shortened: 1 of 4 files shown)


ciao_contrib/criss_cross/criss_cross.py
nplee, line 1: Looking carefully into this file, there is a whole lot of `print` statements rather than using verbosity levels, so even if the user sets `verbose=0`, they will be getting a lot of screen output.  Is this intended, or are they better to change the `print()` to `v1()` so the user's verbosity preference is respected?

It also may be worth considering in the future changing up the printing pattern:

```python
print("\n")
print(f"TIME LOG #{self.counter}  -- {message}")
print(time.asctime(time.localtime()))
print(f"{elapsed} minutes have elapsed")
print("\n")
```

to

```python
print(f"""
TIME LOG #{self.counter}  -- {message}
{time.asctime(time.localtime())}
{elapsed} minutes have elapsed""")
```

for easier readability on the maintenance end.

dprincipe2: I had issues getting the verbosity to work well and it wasn't worth holding up the release at the time. If I recall correctly, the issue was the verbosity controls all the ciao tools in the script and I didn't want things like dmkeypar to output verbosity text at the same level as the rest of crisscross.  The amount of text output now is intended as a compromise to this issue. I'm happy to fix in the future.

